PR #17, opened 2024-02-01
Add utility functions for text processing and visualization

### LINKED ISSUE #9: prompt vis
- merge the efforts from
  - https://github.com/jettjaniak/tinyevals/issues/3 and
  - https://github.com/jettjaniak/tinyevals/issues/4
- make something that works for model comparison, for single model, or without a model

jaidhyani:
This partially addresses #9 (jettjaniak/tinyevals#3)

Including a notebook to demonstrate usage


jaidhyani:
I've added an adapted  version of https://github.com/jettjaniak/tinyevals/pull/25 by @siwei-li in `compare_models.py`

src/delphi/vis/compare_models.py
```diff
@@ -0,0 +1,76 @@
+from dataclasses import dataclass
+
+import torch as t
+import torch.nn as nn
```
jettjaniak: can we just `import torch`?

jettjaniak:
let's move it to delphi/evals/vis.py

src/delphi/vis/vis.py
```diff
@@ -0,0 +1,140 @@
...
+
+    token_class = f"token_{unique_id}"
+    hover_div_id = f"hover_info_{unique_id}"
+
+    for i in range(sample_tok.shape[0]):
+        tok = sample_tok[i].item()
+        data = {}
+        if i > 0:
+            correct_prob = correct_probs[i - 1].item()
+            # ignore type error (tok is a 'Number' by type, but an int in practice)
```
jettjaniak: you could https://docs.python.org/3/library/typing.html#typing.cast ?

src/delphi/vis/vis.py
```diff
@@ -0,0 +1,140 @@
...
+    "margin": "1px 0px 1px 1px",
+    "padding": "0px 1px 1px 1px",
+}
+_token_style_str = " ".join([f"{k}: {v};" for k, v in _token_style.items()])
+
+
+# TODO: basic unit test for visualizations w/ Selenium
+def vis_sample_prediction_probs(
+    sample_tok: Int[t.Tensor, "pos"],
+    correct_probs: Float[t.Tensor, "next_pos"],
```
jettjaniak: if the size of the tensors is supposed to be the same, we should use the same name for the dimension (then it'll fail if they are ever different)

src/delphi/vis/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    hf_ds = load_dataset(f"roneneldan/TinyStories", split=split)
+    dataset = []
+    # hf_ds by type could be an IterableDataset, which is not subscriptable
+    # in practice, it's not
+    for sample_txt in tqdm(hf_ds["text"]):  # type: ignore
+        # encoding issues and rare weird prompts
+        if not set(sample_txt).issubset(ALLOWED_CHARS):
+            continue
+        dataset.append(sample_txt)
+    return dataset
```
jettjaniak: we should get rid of this and make it work with HF Dataset instance

src/delphi/vis/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    probs = get_probs(model, sample_tok)
+    # out of d_vocab values, take the one that corresponds to the correct next token
+    return probs[range(len(probs)), sample_tok[1:]]
+
+
+def get_correct_and_all_probs(model, sample_tok):
+    """Get probabilities for the actual next token and for all predictions"""
+    probs = get_probs(model, sample_tok)
+    correct_probs = probs[range(len(probs)), sample_tok[1:]]
+    return correct_probs, probs
```
jettjaniak: just a note that Victor is using similar utils in #16 - would be good to have them in single, accessible place and tested

jettjaniak:
As mentioned in the original PR on old repo, I don't think it's worth optimizing for performance in the visualization code. It feel it could be much simpler if we dropped the vectorization

src/delphi/vis/utils.py
siwei-li: Should we have a `utils` module to hold different utility functions?

jettjaniak: yeah, delphi/eval/utils.py sounds good

jettjaniak: I added some stuff in #24, please rebase on top of main and extend this

src/delphi/vis/utils.py
jettjaniak: see #25

src/delphi/vis/utils.py
siwei-li: So I tried to use the `get_next_logprobs()` but we'd need a lot of squeeze() and unsqueeze() that way, could we keep these functions for 1-d token sequences just for visualization?

jettjaniak:
You can add a wrapper with squeeze/unsqueeze to utils

src/delphi/vis/utils.py
jaidhyani: Consolidated in the latest update.

jettjaniak:
I think you can use sth like pyproject.toml to configure isort in all 3 instances (pre-commit, CI, vscode), but I'm fine with either solution

src/delphi/eval/compare_models.py
```diff
@@ -0,0 +1,106 @@
...
+import torch.nn as nn
+from jaxtyping import Int
+from transformers import PreTrainedModel
+
+from delphi.eval.utils import get_correct_and_all_probs
+
+
+@dataclass
+class ModelId:
+    model_name: str
```
jettjaniak: that's too much, str is enough

src/delphi/eval/compare_models.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def _pad_start(tensor: torch.Tensor) -> torch.Tensor:
+    value_to_prepend = -1
+    if len(tensor.shape) == 1:
+        return torch.cat((torch.tensor([value_to_prepend]), tensor))
+    else:
+        # input: 2D tensor of shape [seq_len - 1, top_k]
+        pre = torch.full((1, tensor.size()[-1]), value_to_prepend)
+        return torch.cat((pre, tensor), dim=0)
```
jettjaniak: we don't need that

src/delphi/eval/compare_models.py
```diff
@@ -0,0 +1,106 @@
...
+    top_k_b = torch.topk(probs_b, top_k, dim=-1)
+    top_k_a_probs = torch.gather(probs_a, 1, top_k_b.indices)
+
+    next_probs_a = _pad_start(next_probs_a)
+    next_probs_b = _pad_start(next_probs_b)
+    top_k_b_tokens = _pad_start(top_k_b.indices)
+    top_k_a_probs = _pad_start(top_k_a_probs)
+    top_k_b_probs = _pad_start(top_k_b.values)
+
+    comparisons = []
```
jettjaniak: instead of doing _pad_start let the first element of comparisons be None

src/delphi/eval/utils.py
```diff
@@ -42,3 +54,42 @@ def load_validation_dataset(dataset_name: str) -> Dataset:
...
+
+
+def load_text_from_dataset(dataset: Dataset) -> list[str]:
+    text = []
+    for sample_txt in dataset["story"]:
+        # encoding issues and rare weird prompts
+        if not set(sample_txt).issubset(ALLOWED_CHARS):
+            continue
+        text.append(sample_txt)
+    return text
```
jettjaniak: we don't need this

src/delphi/eval/utils.py
```diff
@@ -42,3 +54,42 @@ def load_validation_dataset(dataset_name: str) -> Dataset:
...
+    return correct_probs, probs
+
+
+def get_correct_and_top_probs(
+    model: PreTrainedModel, sample_tok: Int[torch.Tensor, "seq"], top_k: int = 3
+) -> tuple[Float[torch.Tensor, "next_seq"], torch.return_types.topk]:
+    """Get probabilities for the actual next token and for top k predictions"""
+    correct_probs, probs = get_correct_and_all_probs(model, sample_tok)
+    top_k_probs = torch.topk(probs, top_k, dim=-1)
+    return correct_probs, top_k_probs
```
jettjaniak: too many functions
- you don't need get_single_logprobs
- you can rename get_next_logprobs to get_all_and_next_logprobs and make it return both
- make a wrapper get_next_logprobs around get_all_and_next_logprobs that drops the first return value
- then get_next_and_top_probs can call get_all_and_next_logprobs, do squeeze/unsqueeze and apply exp() to get probs from logprobs

src/delphi/eval/vis.py
```diff
@@ -0,0 +1,141 @@
...
+    "font-size": "14px",
+    "color": "black",
+    "background-color": "white",
+    "margin": "1px 0px 1px 1px",
+    "padding": "0px 1px 1px 1px",
+}
+_token_style_str = " ".join([f"{k}: {v};" for k, v in _token_style.items()])
+
+
+# TODO: basic unit test for visualizations w/ Selenium
```
jettjaniak: I don't think we'll do it so can remove this comment

tests/test_vis.py
```diff
@@ -0,0 +1,55 @@
...
+# define a pytest fixture for the tokenized dataset
+@pytest.fixture
+def ds_tok(tokenizer, ds_txt):
+    return [tokenize(tokenizer, txt) for txt in ds_txt]
+
+
+# define a pytest fixture for a tokenized sample
+@pytest.fixture
+def sample_tok(ds_tok):
+    return ds_tok[0]
```
jettjaniak: I think you can safely replace this with a few lines in test_compare_models

tests/test_vis.py
jettjaniak, line 1: this should be tests/eval/test_compare_models.py

src/delphi/eval/utils.py
```diff
@@ -1,9 +1,14 @@
 from collections.abc import Callable
-from typing import cast
+from typing import List, cast
```
jettjaniak: `List` is deprecated, just use `list`

jettjaniak (changes requested):
thanks for making all of these changes!
I suggested a bunch of simplifications

tests/test_vis.py
jaidhyani: This is from aspirations of a lot more test coverage, to enable easy testing across parameter combinations and fixture re-use - but you're right, it is silly when there's just one test and no parameter combinations being tested.

src/delphi/eval/utils.py
jaidhyani: There's a surprising problem with this that I've just spent entirely too much time chasing down: the `topk` object is constructed in the topk function call and is otherwise seemingly impossible to create. The implementation is getting into cpython magic deep in torch, but basically we can't create a new topk object outside of running topk. That means we need to either re-run the topk call, or wreck all of the subsequent topk type hints.

I've tried everything I can think of to get a topk object out of this and spent entirely too much time on it. So: I'm just adding a `remove_batch_dim` flag to `get_next_and_top_k_probs`.

jaidhyani: Ugh, it's actually worse than this. The topk object is run on the logprobs, which is what then ends up in the visualization with no way to change it. So `get_next_and_top_probs` is going to be slightly redundant and just not wrap `get_all_and_next_logprobs` at all.

jaidhyani:
Everything should be addressed.

src/delphi/eval/compare_models.py
```diff
@@ -67,19 +50,18 @@ def compare_models(
     device = model_a.device
     sample_tok = sample_tok.to(device)
 
-    next_probs_a, probs_a = get_correct_and_all_probs(model_a, sample_tok)
-    next_probs_b, probs_b = get_correct_and_all_probs(model_b, sample_tok)
+    probs_a, next_probs_a = get_all_and_next_logprobs_single(model_a, sample_tok)
+    probs_b, next_probs_b = get_all_and_next_logprobs_single(model_b, sample_tok)
```
jettjaniak: Is this probs or logprobs?

src/delphi/eval/utils.py
```diff
@@ -33,12 +29,39 @@ def gather_logprobs(
...
+
+
+def get_next_and_top_k_probs(
+    model: PreTrainedModel, input_ids: Int[torch.Tensor, "seq"], k: int = 3
+) -> tuple[Float[torch.Tensor, "shorter_seq"], torch.return_types.topk,]:
+    all_logprobs, next_logprobs = get_all_and_next_logprobs_single(model, input_ids)
+    all_probs = torch.exp(all_logprobs)
+    next_probs = torch.exp(next_logprobs)
+    top_k = torch.topk(all_probs, k, dim=-1)
+    return next_probs, top_k
```
jettjaniak, line 64: Are you using this function?

jettjaniak (approved):
Great, thank you! Commented on one bug, please make sure to correct it before merging.

src/delphi/eval/utils.py
jaidhyani: In the demo notebook

src/delphi/eval/compare_models.py
jaidhyani: ! Good catch!

PR #18, opened 2024-02-01
Add load_clean_dataset function to dataset.py

jaidhyani:
porting jettjaniak/tinyevals#18 by @menamerai


jettjaniak:
we don't need it anymore

PR #21, opened 2024-02-02
add token labelling

### LINKED ISSUE #12: categorize tokens
https://github.com/jettjaniak/tinyevals/issues/9

joshuawe:


joshuawe:
Are the files that have been added stored in a good location?
I think not.

menamerai:
@joshuawe I think you need to add `spacy` to the `requirements.txt`

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,184 @@
+"""
+This script creates labels for tokens in a sentence. 
+It takes the context of the token into account. 
+Additionally, it can visualize the sentences and their poart-of-speech (POS) tags.
+"""
+
+from pprint import pprint
+from typing import List, Optional
```
jettjaniak: you don't need to import `List` for typing, you can just use `list`

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,184 @@
...
+    else:
+        glossary = spacy.glossary.GLOSSARY
+        print(
+            f"Explanation of all {len(glossary.keys())} token labels (POS, dependency, NER, ...):"
+        )
+        for label, key in glossary.items():
+            print("   ", label.ljust(10), key)
+
+
+def label_single_token(token: Token) -> List:
```
jettjaniak: I would like to have function that just takes a `str` token label. Should we return this as a dict?

jettjaniak (commented):
Let's make a pickle with dict[int, dict[str, bool]] that labels every single token in the tokenizer.

src/delphi/eval/token_labelling.py
joshuawe: You are right.
But unfortunately my venv was created using Python 3.8 where `List` is required.

> In Python 3.8 and earlier versions, to specify the type of elements within containers like lists, you need to import List from the typing module. This is because the built-in list type was not subscriptable (you couldn't use list[int] directly for type hints).

jettjaniak: please update to 3.10, you'll run into issues when we use features from 3.9+ (I think we already do?)

jettjaniak: LMK if you need help on this

joshuawe: Done

joshuawe:
Have a look at the notebook, where function calling as well as the creation of the pickle object which is basically Look Up Table containing a dict with keys *token_id* and value *token_labels*.

Note that *token_labels* is itself a dict with all keys *label_name* and values *bool*

For example:
```
token_ids_lablled = {
  0: {'Starts with space': False, 'Capitalized': True, 'Is Noun': True, 'Is Pronoun': False, 'Is Adjective': False, 'Is Verb': False, 'Is Adverb': False, 'Is Preposition': False, 'Is Conjunction': False, 'Is Interjunction': False, 'Is Named Entity': False},
  1: {'Starts with space': False, 'Capitalized': True, 'Is Noun': True, 'Is Pronoun': False, 'Is Adjective': False, 'Is Verb': False, 'Is Adverb': False, 'Is Preposition': False, 'Is Conjunction': False, 'Is Interjunction': False, 'Is Named Entity': False},
  2: ...
...
}
```

The question that remains is 
1. where to store the pickle. 
2. How to make it accessible. Should it always be loaded into the memory, when the module is imported?

joshuawe:
Another interesting thing is the following, 
the vocab size of the tokenizer is larger than 4000 by far.
```python
>>> from transformers import AutoTokenizer

>>> tokenizer = AutoTokenizer.from_pretrained("roneneldan/TinyStories-1M")

>>> vocab_size = tokenizer.vocab_size
>>> print("The vocab size is:", vocab_size)
50247
``` 

So, am I using the correct tokenizer? And are there multiple tokenizers?

jettjaniak:
The tokenizer is wrong, it's from the original tiny stories models. You need to pass model name from any of the models on delphi-suite Hugging Face to use our small tokenizer that was trained on tiny stories dataset.

requirements.txt
```diff
@@ -9,4 +9,6 @@ black==23.12.1
 jaxtyping==0.2.25
 beartype==0.16.4
 pre-commit==3.6.0
-isort==5.13.2
\ No newline at end of file
+isort==5.13.2
+spacy==3.7.2
+chardet==5.2.0
```
jettjaniak: what is it?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-torch==2.1.2
+# torch==2.1.2
```
jettjaniak: ?

src/delphi/__init__.py
```diff
@@ -1,3 +1,5 @@
 from beartype.claw import beartype_this_package  # <-- hype comes
 
+from . import eval
+
```
jettjaniak: why?

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,211 @@
...
+    #     lambda token: token.dep_ == "ROOT"
+    # ),  # root of the sentence (often a verb)
+    # "Is auxiliary": (lambda token: token.dep_ == "aux"),  # redundant
+    # --- Named entity recognition (NER) categories ---
+    # "Named Entity Type": (lambda token: token.ent_type_),  # '', 'PERSON', 'ORG', 'GPE', ..
+    "Is Named Entity": (lambda token: token.ent_type_ != ""),
+}
+
+
+def explain_Token_labels(token: Optional[Token] = None) -> None:
```
jettjaniak: lowercase please

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,211 @@
...
+        corresponding boolean values.
+    """
+    assert isinstance(token, Token)
+    labels = dict()  #  The list holding labels of a single token
+    for label_name, category_check in TOKEN_LABELS.items():
+        labels[label_name] = category_check(token)
+    return labels
+
+
+def label_sentence(tokens: Union[Doc, list[Token]]) -> list[dict[str, bool]]:
```
jettjaniak:
```suggestion
def label_sentence(tokens: Doc | list[Token]) -> list[dict[str, bool]]:
```
Union is deprecated

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,211 @@
...
+        Whether to print the tokens and their labels to the console, by default False.
+
+    Returns
+    -------
+    list[list[dict[str, bool]]
+        Returns a list of sentences. Each sentence contains a list of its
+        corresponding token length where each entry provides the labels/categories
+        for the token. Sentence -> Token -> Labels
+    """
+    assert isinstance(sentences, list)
```
jettjaniak: Redundant, we're running beartype. Remove it and try passing something different than specified type, it should fail.

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,211 @@
...
+    verbose : bool, optional
+        Whether to print the tokens and their labels to the console, by default False.
+
+    Returns
+    -------
+    list[list[dict[str, bool]]
+        Returns a list of sentences. Each sentence contains a list of its
+        corresponding token length where each entry provides the labels/categories
+        for the token. Sentence -> Token -> Labels
+    """
```
jettjaniak, line 167: My preferred docstring template is 

"""Short one-line description

Optional longer description
"""

If you want to list and describe all arguments that's fine, but specifying their type and optionality is redundant, you can see all of that in function definition.

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # go through each token in the sentence
+        for token, labelled_token in zip(doc, labelled_tokens):
+            # labelled_token = label_single_Token(token)
+            # labels = list()  #  The list holding labels of a single token
+            # for _, category_check in TOKEN_LABELS.items():
+            #     label = category_check(token)
+            #     labels.append(label)
+            # add current token's to the list
+            # labelled_tokens.append(labelled_token)
```
jettjaniak: ?

jettjaniak:
you need to rebase on top of main

jettjaniak:
The notebook is good as an explanation, but generating the labels shouldn't require running it. Let's add something in scripts/ for that.

jettjaniak:
Is there any reason we don't have a function like `label_token(token: str) -> dict[str, bool]`? Combining random tokens into sentences seems opaque and bug prone.

jettjaniak:
for now just store the pickle in eval directory

requirements.txt
joshuawe: `spacy` -> for LLM tasks such as analysis of tokens (part of speach POS)
`chardet` -> a dependency required for `spacy` that was not installed automatically

requirements.txt
joshuawe: ah damn it, sorry, mistake from my side.

src/delphi/__init__.py
joshuawe: I still require this so I can import everything under `delphi`. For WHATEVER reason an empty `__init__.py` file will not allow me to import it. (of course, I used `pip install -e .`)

jettjaniak:
We should debug this. What is your exact setup? System, python interpreter,
how you make your venv etc. Also, can you `import sys; print(sys.path)`

joshuawe:
> We should debug this. What is your exact setup? System, python interpreter, how you make your venv etc. Also, can you `import sys; print(sys.path)`

My setup:
- Windows
- conda for venv creation
- `pip install -e .`
- Python interpreter by running `where Python`, the first entry is : 
  `C:/Users/.../anaconda3/envs/delphi2/python.exe`


Here the other outputs
```Python
>>> print(dir(delphi))
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'beartype_this_package', 'eval']

>>> import sys; print(sys.path)
['c:\\"path to \\delphi\\notebooks', 'c:\\Users\\...\\anaconda3\\envs\\delphi2\\python310.zip', 'c:\\Users\\...\\anaconda3\\envs\\delphi2\\DLLs', 'c:\\Users\\...\\anaconda3\\envs\\delphi2\\lib', 'c:\\Users\\...\\anaconda3\\envs\\delphi2', '', 'c:\\Users\\...\\anaconda3\\envs\\delphi2\\lib\\site-packages', 'c:\\users\\...\\joshua\\research\\2024_01_tinyevals\\delphi\\src', 'c:\\Users\\...\\anaconda3\\envs\\delphi2\\lib\\site-packages\\win32', 'c:\\Users\\...\\anaconda3\\envs\\delphi2\\lib\\site-packages\\win32\\lib', 'c:\\Users\\...\\anaconda3\\envs\\delphi2\\lib\\site-packages\\Pythonwin']
```

jettjaniak:
The only thing that comes to mind is conda being weird. You could just download and install python for windows and make venv this way

joshuawe:
- added tests
- added a script `label_all_tokens.py` that does some token labelling
- now two files are saved under `src/delphi/eval/`. using the tokenizer from a given huggingface language model repo
  1. the tokens and their ids are stored in `all_tokens_{huggingface_model_name}.txt`
  2. The labelled tokens stored as `labelled_token_ids_dict_{huggingface_model_name}.pkl`
- had to update requirements in order to extract/run the delphi tokenizer

Small todo's left for tomorrow:
- [ ] Clear up the `__init__.py` files again
- [x] Implement the function that takes in a token as a `str` and checks for its labels in the token_label_dict
  @jettjaniak Should the tokenized dict be loaded in memory *always*, i.e. when the module is loaded it also loads the pickle?

jettjaniak:
> the tokens and their ids are stored in all_tokens_{huggingface_model_name}.txt
The labelled tokens stored as labelled_token_ids_dict_{huggingface_model_name}.pkl

Can we drop the _{huggingface_model_name}? All delphi models use the same tokenizer.

>  Implement the function that takes in a token as a str and checks for its labels in the token_label_dict
@jettjaniak Should the tokenized dict be loaded in memory always, i.e. when the module is loaded it also loads the pickle?

I assume this question only applies if you're implementing this function? I'd say don't implement it, at least for now. The pickle is enough for the demo. And in this case, you should never be loading the pickle in code included in this PR.

jettjaniak:
branch is out-of-date with main, you need to `git rebase main`

jettjaniak:
were you using `git merge`? something is wrong with "Files changed" tab in this PR and with the commit history

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,198 @@
+"""
+This script creates labels for tokens in a sentence. 
+It takes the context of the token into account. 
+Additionally, it can visualize the sentences and their poart-of-speech (POS) tags.
+"""
```
jettjaniak: This is not a script, but a module - collection of importable functions. They should have their own docstrings, but file docstring is not necessary.

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,198 @@
...
+
+from typing import Callable, Optional
+
+import spacy  # pylint: disable=import-error
+from spacy.tokens import Doc  # pylint: disable=import-error
```
jettjaniak: can you remove the # pylint comments now?

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,198 @@
...
+    token : Token
+        The token to be labelled.
+
+    Returns
+    -------
+    dict[str, bool]
+        Returns a dictionary with the token's labels as keys and their
+        corresponding boolean values.
+    """
+    assert isinstance(token, Token)
```
jettjaniak: beartype should take care of this automatically

src/delphi/eval/token_labelling.py
```diff
@@ -0,0 +1,198 @@
...
+    labels = dict()  #  The list holding labels of a single token
```
jettjaniak: not a list

tests/eval/test_token_labelling.py
```diff
@@ -0,0 +1,114 @@
+import pytest
+import spacy
+from spacy.language import Language
+from spacy.tokens import Doc
+
+import delphi.eval.token_labelling as tl
+
+
+@pytest.fixture
+def create_dummy_doc() -> tuple[str, Doc, dict[str, bool]]:
```
jettjaniak: fixture should probably be named dummy_doc() instead of create_dummy_doc()

tests/eval/test_token_labelling.py
jettjaniak, line 1: nice!

scripts/label_all_tokens.py
```diff
@@ -0,0 +1,116 @@
...
+def decode(
+    tokenizer: PreTrainedTokenizer | PreTrainedTokenizerFast, token_ids: list[int]
+) -> str:
+    return tokenizer.decode(token_ids, skip_special_tokens=True)
+
+
+def main():
+    print("\n", " LABEL ALL TOKENS ".center(50, "="), "\n")
+    # Access command-line arguments
+    args = sys.argv[1:]
```
jettjaniak: please use argparse

scripts/label_all_tokens.py
```diff
@@ -0,0 +1,116 @@
...
+    print(f"You chose the model: {model_name}\n")
+    print(
+        f"The language model will be loaded from Huggingface and its tokenizer used to do two things:\n\t1) Create a list of all tokens in the tokenizer's vocabulary.\n\t2) Label each token with its part of speech, dependency, and named entity recognition tags.\nThe respective results will be saved to files located at: '{SAVE_DIR}'\n"
+    )
+
+    # ================ (1) =================
+    print("(1) Create a list of all tokens in the tokenizer's vocabulary ...")
+
+    # Load the tokenizer from Huggingface
+    tokenizer = AutoTokenizer.from_pretrained(model_name)
```
jettjaniak, line 52: is this the line that requires sentencepiece and protobuf?

scripts/label_all_tokens.py
```diff
@@ -0,0 +1,116 @@
...
+    batch_size = 500
+    # we iterate (batchwise) over all token_ids, individually takes too much time
+    for start in tqdm(range(0, max_token_id, batch_size), desc="Labelling tokens"):
+        # create a batch of token_ids
+        end = min(start + batch_size, max_token_id)
+        token_ids = list(range(start, end))
+        # decode the token_ids to get a list of tokens, a 'sentence'
+        tokens = decode(tokenizer, token_ids)  # list of tokens == sentence
+        # put the sentence into a list, to make it a batch of sentences
+        sentences = [tokens]
```
jettjaniak: This seems wrong. Are you turning a list of unrelated tokens into a single string?

jettjaniak: How are they being split afterwards?

jettjaniak:
did you try venv w/o conda?

scripts/label_all_tokens.py
joshuawe: yes exactly

scripts/label_all_tokens.py
joshuawe: True, it might not make sense to process the tokens batch-wise.
Doing it individually is extremely slow. And as we are NOT looking at dependencies of tokens in a sentence (subject, object, etc.) I assumed it would be fine to increase the computational speed this way.
I had also assumed that we might call this function more frequently.
But as we will use the look-up-dict I can calculate it once but slowly instead. (So not batched)

joshuawe:
ok, everything is set up.
It can definitely be used for demo purposes.
Please checkout the notebook.

What I just realised is that there are a few tokens that seem not to be labelled entirely correctly. I will check this in an upcoming task.

jaidhyani:
@joshuawe - would it be okay if I adapted some of this code for the demo without pulling in the whole PR?

joshuawe:
@jaidhyani  you can do that, if you think that makes sense.
But why not integrate the PR?
The reason for the previously failed check was that it wants to load a model that is too for the github action. I chose a smaller one, so the check passes. So, if that was the previous aprehension to not merge the PR, it should now be resolved.

jettjaniak (approved):

jettjaniak:
rebase required

PR #22, opened 2024-02-02
Ported token mapping function

### LINKED ISSUE #8: map tokens to prompts and positions
https://github.com/jettjaniak/tinyevals/issues/7

menamerai:
Ported mapping function over from previous repo. Made modifications so it works with the HF Dataset classes.

For issue #8.


scripts/map_tokens.py
```diff
@@ -0,0 +1,14 @@
+from datasets import DatasetDict, IterableDataset, IterableDatasetDict
+from datasets.arrow_dataset import Dataset
+
+
+def token_map(
+    tokenized_dataset: DatasetDict | Dataset | IterableDatasetDict | IterableDataset,
```
jettjaniak: why so many?

jettjaniak (changes requested):
so for this to be a script it has to do something, right? Like take dataset name as argument and save/upload the output. I think I'm satisfied with this being simple enough to not require tests, but we need to run it and confirm it works

scripts/map_tokens.py
menamerai: The `load_dataset` function returns one of those 4 types. Pylance got mad at me if I just use one of the four, so I included all 4 possible return types here. I think our tokenized dataset only has the `Dataset` class.

jettjaniak: You can try using https://docs.python.org/3/library/typing.html#typing.cast let pyright / pylance know we will be only using Dataset only before calling this function

src/delphi/dataset/token_map.py
```diff
@@ -6,6 +6,15 @@
...
 ) -> dict[int, list[tuple[int, int]]]:
+    """Return a mapping of tokens to their (prompt_idx, token_idx) locations in the tokenized_dataset.
+
+    Args:
+        tokenized_dataset (Dataset): A tokenized dataset.
+
+    Returns:
+        dict[int, list[tuple[int, int]]]: A mapping of tokens to their (prompt_idx, token_idx)
+            locations in the tokenized_dataset.
+    """
```
jettjaniak: first line of the docstring is enough, everything else is redundant

src/delphi/dataset/token_map.py
```diff
@@ -0,0 +1,25 @@
...
+    mapping = {}
+    tokenized_dataset = cast(Dataset, tokenized_dataset)
```
jettjaniak: Why cast here? What happens if you remove it?

src/delphi/dataset/token_map.py
```diff
@@ -0,0 +1,25 @@
...
+    for prompt_idx, prompt in enumerate(tokenized_dataset):
+        prompt = cast(dict, prompt)
```
jettjaniak: what's the error here, pyright doesn't realize prompt has a __get_item__?

tests/test_dataset.py
```diff
@@ -0,0 +1,73 @@
...
+                    3,
+                    4,
+                    5,
+                    0,
+                    6,
+                    7,
+                ],
+            ]
+        }
+    )
```
jettjaniak: this is bad xD
you can add a noqa comment for black and keep it at reasonable number of lines

jettjaniak:
test looks good!
left some minor comments
please move token_map.py to eval folder and the test to tests/eval/test_token_map.py
then we need a script in scripts/ that takes HF dataset name and produces a pickle
click "ready for review" when that's done and ping me please

src/delphi/dataset/token_map.py
menamerai: Okay, will remove the other lines!

src/delphi/dataset/token_map.py
menamerai: I initially cast in hope of getting pyright to realize prompt has a get_item, but the problem is elsewhere. Deleting this line seems to not affect the function in any way, so I'll remove it

src/delphi/dataset/token_map.py
menamerai: Yes. this is the full message:
```
Argument of type "Literal['tokens']" cannot be assigned to parameter "__s" of type "slice" in function "__getitem__"
  "Literal['tokens']" is incompatible with "slice"
```
I tested the dataset live, and the prompt seems to be of type dict when I access it, so I casted it to dict. Should I get a typeignore comment in or do something else?

tests/test_dataset.py
menamerai: Hahaha yeah thank you I wasn't sure how to deal with that

menamerai:
@jettjaniak I've updated most things, but still a little confused about some things:
- What should I do about the typing on the prompt? Should I ignore the type with `# type: ignore`, should I keep the cast, or should I do something else?
- I only added a basic script that can take any hf dataset string and make a pickle in `data/pickle_name.pkl`, but should it be more specific? Should it do something else?

jettjaniak:
let's discuss at check-in

jettjaniak (approved):

PR #23, opened 2024-02-03
Permute list in place

### LINKED ISSUE #11: deterministic permutation function
- context
  - we want reproducible training
  - one of the elements of that is ensuring model is seeing samples in the same order in every run
  - we need to shuffle samples in each epoch
- we will go with a deterministic function in Python
  - `permute(input_output: list, seed: int) -> None`
    - it returns nothing, but permutes the list in-place
    - permutation doesn't depend on the elements
    - but only on the seed and length of the input
    - put it in train/permute.py
- we don't want to use any libraries, as RNG engines in these [can change over time](https://docs.python.org/3/library/random.html#notes-on-reproducibility)
- here's a (buggy) idea how to do it https://chat.openai.com/c/b4c2a457-1fb3-4f20-8016-6bdf71a0a2d3
  - use some reasonable LCG parameters (ideally from an old classic implementation)

SrGonao:
I think this is good


src/delphi/train/permute.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+
+    def __init__(self, seed):
+        self.state = seed
+
+    def set_seed(self, seed):
+        """
+        Set the seed
+        """
+        self.state = seed
```
jettjaniak: redundant?

src/delphi/train/permute.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def permute(self, list):
+        """
+        Permute a list in place using Fisher-Yates shuffle
+        """
+        for i in range(len(list)):
+            j = int(self.random_number_generator() * (i + 1))
+            # self.state = self.state + 1
```
jettjaniak: remove

src/delphi/train/permute.py
```diff
@@ -0,0 +1,36 @@
...
+            list[i], list[j] = list[j], list[i]
+        return list
+
+    def random_number_generator(self):
+        """
+        Generate a random number between 0 and 1 using minstd_rand from C++11
+        """
+        self.state = (self.state * 48271) % 4294967296
+        return self.state / 4294967296
```
jettjaniak: let's just return the state as an int
can you add type hint for return btw?

src/delphi/train/permute.py
```diff
@@ -0,0 +1,36 @@
...
+        Set the seed
+        """
+        self.state = seed
+
+    def permute(self, list):
+        """
+        Permute a list in place using Fisher-Yates shuffle
+        """
+        for i in range(len(list)):
+            j = int(self.random_number_generator() * (i + 1))
```
jettjaniak: if random_number_generator() would return int then you just need to % i (or i+1?)

tests/test_permutations.py
```diff
@@ -0,0 +1,20 @@
+from delphi.train.permute import Permutation
```
jettjaniak: I would move the file to tests/train/test_permute.py

tests/test_permutations.py
```diff
@@ -0,0 +1,20 @@
...
+        assert perm.random_number_generator() >= 0
+        assert perm.random_number_generator() <= 1
+    state = perm.state
+    assert state == 822144001
+    # Test the permutation
+    test_list = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+    length = len(test_list)
+    for i in range(100000):
+        perm.permute(test_list)
+        assert len(test_list) == length
```
jettjaniak: you could check if it's a permutation, i.e. all elements are present before and after

jettjaniak (changes requested):
Thanks, looks good overall!

I think we want RNG as a class, but the permutation should just be a standalone function that makes an instance of RNG, does the permutation and discards of the RNG.

Another thing to figure out is how to make this function take 2 arguments: seed and epoch number.

Also, CI is failing on formatting, see https://github.com/delphi-suite/delphi?tab=readme-ov-file#formatting

src/delphi/train/permute.py
SrGonao: I changed this to random uniform and created a generate_number, which generates a number from 0 to max (which was required for the correct shuffling algorithm)

src/delphi/train/permute.py
SrGonao: I don't understand why it should return the state, I refactored it in a different way

SrGonao:
I resolved most of your changes but there's some where I'm not sure what you meant,
I assumed that the by the epoch number you wanted to be able to say a epoch number and go to the state of the RNG as if you had made epoch number -1 epochs. (I have a commit that I didn't push about starting the epoch at 1 and not 0, not sure what you think)

jettjaniak (approved):

PR #25, opened 2024-02-08
eval.utils.load_validation_dataset

### LINKED ISSUE #4: load validation dataset only
- HF load_dataset() will download whole repo even if we only want validation which is 1% of it
- I'd like to have a function that will download only validation files and return HF dataset object
- a relevant sample of code I was using before

```
def get_ds_shard_list(dataset_name: str, shard_nr: int, n_total_shards: int):
    print(f"Loading shard {shard_nr} from {dataset_name}...")
    data_str = f"data/train-{shard_nr:05}-of-{n_total_shards:05}-*.parquet"
    return load_dataset(f"NeelNanda/{dataset_name}", data_files=data_str, verification_mode='no_checks', split="train")
```

- `data_str = f"data/validation-*.parquet"` should do it?
- test it on jbrinkma/tinystories-v2-clean and jbrinkma/tinystories-v2-clean-tokenized

jettjaniak:


src/delphi/eval/utils.py
```diff
@@ -25,3 +27,18 @@ def get_next_logprobs(
...
+    data_str = f"data/validation-*.parquet"
+    dataset = load_dataset(
+        dataset_name,
+        data_files=data_str,
+        verification_mode="no_checks",
+        # this seems to be the only split when using data_files
+        # regardless of the files we're actually loading
+        split="train",
+    )
+    return cast(Dataset, dataset)
```
siwei-li, line 44: One question: we are still add a BOS token to every sequence in `tokenized = load_validation_dataset("tinystories-v2-clean-tokenized-v0")` before batching, right?

jettjaniak: No, we don't prepend anything in the tokenized dataset. -v0 is weird and doesn't have BOS, but this is what the v0 models were trained on. Final version will have BOS

PR #26, opened 2024-02-09
Add model/dataset cache to GH actions

### LINKED ISSUE #1: use cache in github actions
right now pip is downloading and installing requirements each time which takes about 2 minutes, we could speed it up https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

more importantly, if we want to use models and datasets in our tests, we should cache them as well
see here for how it can be done https://github.com/neelnanda-io/TransformerLens/blob/53c95f3f154b2933dcc84647bd80bf29a9e4e2d2/.github/workflows/checks.yml#L58-L65

jaidhyani:
This PR enables caching for huggingface models/datasets (#1). It does require users to manually increment a cache identifier when we want to change what's going into the cache, since I couldn't think of a good way to automatically check what will be needed in the pipeline ahead of time. Instructions have been added to README.md.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -21,6 +21,7 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.10"
+          cache: "pip"
       - name: cache models and datasets
         uses: actions/cache@v3
         with:
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -21,6 +21,12 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: "3.10"
+      - name: cache models and datasets
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/huggingface
+          key: ${{ runner.os }}-huggingface-cache-v1  # increment this key to invalidate the cache when new models/datasets are added
       - name: dependencies
         run: |
           python -m pip install --upgrade pip
--- a/README.md
+++ b/README.md
@@ -1,13 +1,16 @@
 # Delphi
+
 Interpreting Small Language Models Across Time and Scale
 
 # setup
+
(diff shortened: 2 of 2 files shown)


jettjaniak:
Amazing, thanks! Can you cache dependencies as well? Currently they're the bottleneck I believe.

jettjaniak:
bizarre that README got formatted by something. Did you use something else than text editor?

jaidhyani:
Ah, I think it was [Prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode). I've disabled it for this workspace.

jaidhyani:
> Amazing, thanks! Can you cache dependencies as well? Currently they're the bottleneck I believe.

Added via the integrated caching functionality in `setup-python`. This *should* just work, testing it now.

jaidhyani:
So good news/bad news:

Good news: This does work, pip packages are saved and don't need to be downloaded again. [First run writes to cache](https://github.com/delphi-suite/delphi/actions/runs/7859272309/job/21445183815?pr=26), [second run reads from cache](https://github.com/delphi-suite/delphi/actions/runs/7859272309/job/21445257436?pr=26).

Bad news: This only saves about 30 seconds, most of the time is spent on the actual install

Good news: It's theoretically possible to save the post-install state, see https://archive.ph/WAex2

Bad news: This potentially breaks a lot of integrations by Doing Things In An Unexpected Way. And discussion of the relevant open issue on the action itself suggests that people haven't really gotten too much benefit out of it: https://github.com/actions/setup-python/issues/330

Tentative conclusion: Take the small win and revisit later if and only if this is enough of a blocker that we're willing to dedicate a lot of person-hours to improving it instead of working on something else.

jettjaniak (approved):
agreed with the conclusion, thanks!

PR #28, opened 2024-02-11
vscode now uses black from venv instead of bringing its own and causing conflicts

### LINKED ISSUE #30: Fix black version conflicts

jaidhyani:


jettjaniak (approved):
I think I checked and the plugin page was saying it will use fromEnvironment as long as black is installed? Weird.

jaidhyani:
> I think I checked and the plugin page was saying it will use fromEnvironment as long as black is installed? Weird.

I'm betting I didn't have black installed in the default environment when I first installed the extension.

PR #32, opened 2024-02-14
End-to-end evaluation demo

### LINKED ISSUE #13: first end2end eval demo
- just loss over groups of tokens on a few models
- interactive plotly line plot (you can choose tokens group with a drop down)
- models `jbrinkma/delphi-llama2-{100,200,400}k`
- dataset: validation split of `jbrinkma/tinystories-v2-clean-tokenized`

image

- depends on: #12 #10 #8 #4 #5
- notebook using some library functions is fine
  - prioritize making a functional plot above clean code, tests etc
  - when it works we can make it pretty

jaidhyani:
WIP


jaidhyani:
image

jaidhyani:
image

jaidhyani:
This ended up being an embarrassing quantity of iteration and written-then-deleted code for what was ultimately a pretty small PR. On the upside, I feel like I gained a better understanding of a lot of the codebase in the process.

data/README.md
jettjaniak, line 1: move?

scripts/map_tokens.py
menamerai, line 1: I think you might need to change this output to be saved in the static dir

data/model_group_stats.pkl
menamerai, line 1: Should be removed and added to static

data/token_map.pkl
menamerai, line 1: Should be removed and added to static

menamerai (changes requested):

scripts/map_tokens.py
jaidhyani: Yeah, definitely. I just didn't want to scope creep this PR, and or make a hard dependency on the static dir PR.

data/README.md
jaidhyani: Yup, now that the static dir PR is in

src/delphi/eval/utils.py
```diff
@@ -87,3 +89,16 @@ def tokenize(
         Int[torch.Tensor, "seq"],
         tokenizer.encode(tokenizer.bos_token + sample_txt, return_tensors="pt")[0],
     )
+
+
+def load_logprob_dataset(model: str):
+    return cast(
+        Dataset, load_dataset(f"transcendingvictor/{model}-validation-logprobs")
+    )
```
jettjaniak: here it would probably be sufficient to have return type hint and suppress pyright?

src/delphi/eval/utils.py
```diff
@@ -87,3 +89,16 @@ def tokenize(
...
+
+
+def load_logprob_datasets(split: str = "validation") -> dict[str, list[list[float]]]:
+    return {
+        model: cast(dict, load_logprob_dataset(model)[split])["logprobs"]
```
jettjaniak, line 100: I'm worried how much we have to use casting with Dataset objects :/

data/README.md
```diff
@@ -0,0 +1,10 @@
+# TODO: move this to delphi/static
+# Static Data Files
+
+
+## `token_map.pkl`
```
jettjaniak: that lives in HF now

requirements.txt
```diff
@@ -13,4 +13,6 @@ isort==5.13.2
 spacy==3.7.2
 chardet==5.2.0
 sentencepiece==0.1.99
-protobuf==4.25.2
\ No newline at end of file
+protobuf==4.25.2
+plotly==5.18.0
+spacy-transformers==1.3.4
```
jettjaniak, line 18: what is this?

src/delphi/eval/calc_model_group_stats.py
```diff
@@ -0,0 +1,54 @@
+import numpy as np
+
+
+def calc_model_group_stats(
+    tokenized_corpus_dataset: list,
+    logprob_datasets: dict[str, list[list[float]]],
+    token_groups: dict[int, dict[str, bool]],
```
jettjaniak: What we used to do at my old job with dicts was sth like
```suggestion
    logprobs_by_model: dict[str, list[list[float]]],
    labels_by_token: dict[int, dict[str, bool]],
```
maybe these aren't the best examples / names, but the idea was to describe what the key in a dict was. Sometimes recursively like x_by_y_by_z for a nested dict

src/delphi/eval/calc_model_group_stats.py
```diff
@@ -0,0 +1,54 @@
...
+    token_labels: list[str],
+) -> dict[tuple[str, str], dict[str, float]]:
```
jettjaniak, line 9: we should probably start using more dataclasses

jettjaniak (approved):
I'm happy with this. Left a few comments, use your best judgement to decide what to address and what to ignore for now

menamerai (approved):

src/delphi/eval/calc_model_group_stats.py
jaidhyani: That's a nice convention, I'm in.

src/delphi/eval/calc_model_group_stats.py
jaidhyani: Strong endorse

src/delphi/eval/utils.py
jaidhyani: Yeah, the dataset library clearly wasn't designed with type-guarantees in mind. Building a library of functions that take care of the casting and other stuff we don't want to worry about in our day-to-day might be a good idea. Ticket would be something like "identify repetitive casts in the codebase and replace them with library functions".

jaidhyani (commented):
> what is this? 

spacy-transformers is a dependency I needed to get the token labeling code to work that I think just got missed by accident earlier (it's a hidden dependency that only shows up at runtime).

requirements.txt
jaidhyani: Sorry, meant to reply to this but the vscode inline thingy is kind of buggy and it ended up as a separate comment. Anyway:

spacy-transformers is a dependency I needed to get the token labeling code to work that I think just got missed by accident earlier (it's a hidden dependency that only shows up at runtime).

PR #35, opened 2024-02-15
performance plot

### LINKED ISSUE #5: interactive performance plot
https://github.com/jettjaniak/tinyevals/issues/11

siwei-li:


src/delphi/dataset/mock_per_token_performance.py
```diff
@@ -0,0 +1,47 @@
+from math import log2
+from random import random, seed
```
jettjaniak: unless importing from our own library, default to importing modules instead of specific functions / classes
```suggestion
import math
import random
```

src/delphi/dataset/mock_per_token_performance.py
```diff
@@ -0,0 +1,47 @@
...
+    },
+    "llama2-1b": {
+        "nouns": [log2(random()) for _ in range(200)],
+        "verbs": [log2(random()) for _ in range(100)],
+        "adjectives": [log2(random()) for _ in range(250)],
+        "adverbs": [log2(random()) for _ in range(40)],
+        "pronouns": [log2(random()) for _ in range(55)],
+        "prepositions": [log2(random()) for _ in range(234)],
+    },
+}
```
jettjaniak: you could make this 5x shorter with a for loop

src/delphi/dataset/mock_per_token_performance.py
jettjaniak, line 1: Let's make a notebook that showcases the usage of our plotting function and use this mock data there. Then remove this file.

src/delphi/eval/vis_per_token_model.py
```diff
@@ -0,0 +1,37 @@
+import ipywidgets as widgets
+import matplotlib.pyplot as plt
+import numpy as np
+from ipywidgets import interact
```
jettjaniak:
```suggestion
import matplotlib.pyplot as plt
import numpy as np
import ipywidgets
```

src/delphi/eval/vis_per_token_model.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def visualize_per_token_category(input):
```
jettjaniak: add type hints

src/delphi/eval/vis_per_token_model.py
```diff
@@ -0,0 +1,37 @@
...
+    model_names = list(input.keys())
+    categories = list(list(input.values())[0].keys())
```
jettjaniak:
```suggestion
    categories = list(input[model_names[0]].keys())
```

src/delphi/eval/vis_per_token_model.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def _f(category):
```
jettjaniak: more descriptive name, underscore unnecessary (it's already defined inside another function, so you can't import it)

jettjaniak: type hints

src/delphi/eval/vis_per_token_model.py
```diff
@@ -0,0 +1,37 @@
...
+            placeholder="",
+            description="Token Category:",
+            disabled=False,
+        ),
+    )
+
+
+# Usage:
+# from dataset.mock_per_token_performance import performance_datas
+# visualize_per_token_category(performance_data)
```
jettjaniak: we probably want a notebook instead, but that would be a better approach for the future
```suggestion
if __name__ == "__main__":
    from dataset.mock_per_token_performance import performance_datas
    visualize_per_token_category(performance_data)
```

jettjaniak (dismissed):
it's good, let's continue working on it!

1. make plotly plot instead - I'm not sure if using ipywidgets will be the best solution in this case, as plotly has some built-in support for dropdowns, see original issue description
2. change the input format to a tuple of (err_low, value, err_high) instead of list of all values - someone else will take care of computing this efficiently on the dataset scale
3. make it look somewhat like that, with colors as arguments with reasonable defaults; loss is log scale
image
4. see code comments

siwei-li:
I have changed the plotting by using Plotly, the only thing I'm not sure about is this coloring thing @jettjaniak 
> 3. make it look somewhat like that, with colors as arguments with reasonable defaults; loss is log scale

siwei-li:
Now the visualization makes Plotly plot instead -- the only thing I'm unsure about is this coloring thing:
> 3. make it look somewhat like that, with colors as arguments with reasonable defaults; loss is log scale

jaidhyani:
lgtm, nice work

image

jaidhyani (approved):

jaidhyani:
Two things to follow up on:

1. I think we want the loss (y axis) to be log scale.

2. We should make `color` an argument of `visualize_per_token_category` (it can default to "purple")

siwei-li:
Yeah I just thought that we already have data as in logprobs, so the loss calculated is already in log
> 1. I think we want the loss (y axis) to be log scale.

jaidhyani:
Yeah, I think you're right. Jett, thoughts? You mentioned wanting a log
scale in an earlier comment, were you just referring to charting logprob?

src/delphi/eval/vis_per_token_model.py
```diff
@@ -0,0 +1,50 @@
+import ipywidgets
+import numpy as np
+import plotly.graph_objects as go
+from beartype.typing import Dict
```
jettjaniak: you can just use `dict`

jettjaniak:
just had a meeting with Sil, a few small bits to address and we can merge

siwei-li:
Todos:
- take log/linear kwarg options
- coloring kwargs fo the graph
- making dots larger

PR #40, opened 2024-02-18
34 manual token labeling

### LINKED ISSUE #34: manual token labeling
@jaidhyani has more context

joshuawe:
Added a *.CSV file that is cerated during the labelling process of tokens.
While the pickle file (which still exists) has the structure of `dict[token_id, dict[label_name, label_value]]` the CSV now has the form

| token_id | Starts with space | Capitalized | Is Adjective | Is Adposition | ... |
|--------|--------|--------|--------|--------|--------|
| 0 | False | True| False | False | ... |
| 1 | False | False | True| False | ... |
| 2 | False | False | False | False |  ... |
| ... | ... | ... | ... | ... | ... | ... | ... |


scripts/label_all_tokens.py
```diff
@@ -103,6 +104,34 @@ def main():
...
+        assert (
+            label_pandas == label_dict
+        ), f"The dataframes are not equal for row {token_id}\n{label_pandas}\n{label_dict}"
+    print(" completed.")
+
+    # save the dataframe to a csv
+    filename = "labelled_token_ids_df.csv"
+    filepath = SAVE_DIR / filename
+    df.to_csv(filepath, index=False)
+    print(f"Saved the labelled tokens as CSV to:\n\t{filepath}\n")
```
jaidhyani: We should probably make the path a script argument, and not default to outputting data files in code. This isn't a blocker though, we can adjust it later.

jaidhyani:
This is looking good. Here's what I think we want to do to finalize it:

1. Factor out the core token-labeling functionality into the `token_labelling.py` (basically lines 49-88). This probably looks like a function that takes a tokenizer as an argument and returns a `labelled_token_ids_dict` (or similar data structure that fully specifies token labels)

2. Adapt the current script into an `export_labels` script with three arguments: (1) model/tokenizer (as exists in the current script) (2) output format (csv or pkl) and (3) output path. Output path should be a mandatory argument without defaults. This will call the function defined in (1) and basically just handle the file writing.

3. Back in `token_labeling.py`, add a function `import_token_labels` which takes a path argument (to a csv) and returns a `labelled_token_ids_dict`-like data structure specifying token labels.

4. Move generated data files to `delphi/static/` (see currently-a-draft PR #41)

jettjaniak:
Nice! The CSV needs a column with token label though

scripts/label_all_tokens.py
```diff
@@ -32,12 +33,28 @@ def main():
         default="delphi-suite/delphi-llama2-100k",
         required=False,
     )
+    parser.add_argument(
+        "--save_dir", type=str, help="Directory to save the results.", required=True
```
jettjaniak:
```suggestion
        "--save-dir", type=str, help="Directory to save the results.", required=True
```

scripts/label_all_tokens.py
```diff
@@ -32,12 +33,28 @@ def main():
...
+    )
+    parser.add_argument(
+        "--output_format",
```
jettjaniak:
```suggestion
        "--output-format",
```

joshuawe:
> Nice! The CSV needs a column with token label though

@jettjaniak Could you elaborate what you mean?
Would you like to have a header line, that explains the meaning of each column?

jaidhyani:
This is looking good - nice.

src/delphi/eval/token_labelling.py
```diff
@@ -208,3 +213,111 @@ def label_batch_sentences(
...
+    ----------
+    path : str | Path
+        The path to the file.
+
+    Returns
+    -------
+    dict[int, dict[str, bool]]
+        Returns the labelled tokens dict. Each token_id has its own dict having the labels.
+    """
+    global labelled_token_ids_dict
```
jettjaniak: why is this global?

tests/eval/test_token_labelling.py
```diff
@@ -112,3 +117,75 @@ def test_label_batch_sentences(dummy_doc):
...
+    vocab_size = tokenizer.vocab_size
+
+    tokens_str, labelled_token_ids_dict = tl.label_tokens_from_tokenizer(tokenizer)
+    # count the number of lines in the token_str
+    assert tokens_str.count("\n") == (vocab_size + 1)  # + 1, because of token '\n'
+    assert len(labelled_token_ids_dict.keys()) == vocab_size
+    assert is_valid_structure(labelled_token_ids_dict) == True
+
+
+@pytest.mark.dependency(depends=["test_label_tokens_from_tokenizer"])
```
jettjaniak: what does it do?

scripts/label_all_tokens.py
```diff
@@ -26,17 +27,34 @@ def main():
...
+        help="Format to save the results in. Options: csv, pkl. Default: csv.",
+        default="csv",
+        required=False,
+    )
     args = parser.parse_args()
 
     # Access command-line arguments
-
+    # Directory to save the results
+    SAVE_DIR = Path(args.save_dir)
```
jettjaniak: uppercase is generally reserved for constants

jettjaniak:
I don't see a use case for *.pkl anymore

tests/eval/test_token_labelling.py
joshuawe: This decorator has another test run first (the test that it depends on), whose results can be used in another test.
So here, we first create a dict with the token labels in another test and use those results to test subsequent functions of our library.

src/delphi/eval/token_labelling.py
joshuawe: thanks for catching that. Needed the global var during testing, not in the library code

scripts/label_all_tokens.py
joshuawe: made it lower case

joshuawe:
Resolved two comments of yours.

Should I get rid of the *.pkl option entirely before merging? @jettjaniak

tests/eval/test_token_labelling.py
```diff
@@ -112,3 +117,75 @@ def test_label_batch_sentences(dummy_doc):
...
+                f"Main structure is dict, but its keys are either not int or its values are not dicts. Instead key is type {type(key)} and value is type {type(value)}"
+            )
+            return False
+        for sub_key, sub_value in value.items():
+            if not isinstance(sub_key, str) or not isinstance(sub_value, bool):
+                print(
+                    f"The structure dict[int, dict[X, Y]] is True, but either X is not str or Y is not bool. Instead X is type {type(sub_key)} and Y is type {type(sub_value)}"
+                )
+                return False
+    return True
```
jettjaniak, line 146: beartype is doing this automatically

tests/eval/test_token_labelling.py
jettjaniak: I see. I don't think this is a good practice, I suggest you use pytest fixtures instead (built-in, don't require additional packages) https://docs.pytest.org/en/6.2.x/fixture.html

jettjaniak:
I think #47 and #51 should go with this PR
And please add the CSV to static when ready

tests/eval/test_token_labelling.py
joshuawe: In general I disagree about good practice. In bigger projects with large amounts of tests, you would not want to waste compute on tests that anyways depend on other tests. 

Here we do not have a large test set, so if you prefer I can go with fixtures and have the same code from `test_label_tokens_from_tokenizer` a second time.

tests/eval/test_token_labelling.py
joshuawe: Ha, nice. But weirdly enough I needed this to catch a bug because beartype did not throw an error when the type was `dict[int, dict[str, np.array[bool]]`

joshuawe:
- [x] get rid of pickling
- [ ] add str repr into *.csv
  - [x] correctly create & save CSV
  - [ ] correctly load the CSV
  - [ ] Merge the code into this branch of #40
- [x] rewrite testing to not include dependency (slow version OR @pytest.skip)
- [ ] Add the token labelling CSV once it is created to STATIC
- [ ] comment / todo / fixme on export to csv eating spaces

jettjaniak:
- check what's wrong with beartype and nested structures https://beartype.readthedocs.io/en/latest/eli5/
- implement slow version of tests and @pytest.skip or do fixture
- comment / todo / fixme on export to csv eating spaces

jettjaniak:
@transcendingvictor your tasks is blocked by this and Josh won't be available until weekend. Could you complete the items in the comment above and merge this? Hit me up on Discord if sth is unclear

tests/eval/test_token_labelling.py
jettjaniak, line 1: this file should be renamed to spacy_* as well

tests/eval/test_token_labelling.py
```diff
@@ -1,11 +1,17 @@
+import pickle
+from pathlib import Path
+
 import pytest
-import spacy
 from spacy.language import Language
 from spacy.tokens import Doc
+from transformers import AutoTokenizer
 
 import delphi.eval.token_labelling as tl
```
jettjaniak: How did you rename the files? If you do right click and rename in VSCode it should also rename all references

tests/eval/test_token_labelling.py
```diff
@@ -1,11 +1,17 @@
...
 
+labelled_token_ids_dict: dict[int, dict[str, bool]] = {}
+
 
+@pytest.skip("These tests are slow")
 @pytest.fixture
 def dummy_doc() -> tuple[str, Doc, dict[str, bool]]:
```
jettjaniak: this is a fixture, not a test, so you probably don't have to skip it

tests/eval/test_token_labelling.py
```diff
@@ -1,11 +1,17 @@
...
 from spacy.language import Language
 from spacy.tokens import Doc
+from transformers import AutoTokenizer
 
 import delphi.eval.token_labelling as tl
 
+labelled_token_ids_dict: dict[int, dict[str, bool]] = {}
+
 
+@pytest.skip("These tests are slow")
```
jettjaniak: I didn't realize how many there are and that they're all in a single file. I believe you can just do pytest.skip() on the top level, just after the imports to skip the whole file. The reason is "tests are slow and we're not using this module currently"

jettjaniak (approved):

joshuawe:
Thanks @transcendingvictor  for doing this

PR #41, opened 2024-02-19
Added static file folder

### LINKED ISSUE #36: static assets in python package
@jaidhyani has more context

menamerai:
Is there a way to default to using the `delphi/static` folder instead of the `delphi` folder from the `from importlibs.resources import files` @jaidhyani ? In here I just join the path with `static` every time.


jaidhyani:
Sort of - `files` takes a module (or string-specifying-a-module) as an argument. So to make it point somewhere other than our package, we would need to give it a different argument. One thing we could do would be to add an `__init__.py` to the `static` dir, and then we could do `files(delphi.static)`. Either approach is fine, I think - your call.

But either way, you're right that we shouldn't require people to type that out each time. Let's add a `delphi/constants.py`, which for now can just contain `STATIC_ASSETS_DIR` definition.

menamerai:
@jaidhyani I added `__init__.py` to the directory, and added the constants file. How does it look now?

jaidhyani (approved):
This looks good to me. Minor follow-up tasks (edit: added as issue #42). These aren't blockers, code can be merged

1. We shouldn't necessarily force the output into the `STATIC_ASSETS_DIR`. Especially if the user specified a full path or a partial path with directory references, e.g.:
 
 `/home/me/Desktop/tokens.csv`
 `subdir/tokens.csv`
 
In the first case, the user has specified a full path, and we should use it. In the second, we can assume they're referring to a relative path from the current working directory, and we should output there.
 
We can get that behavior by calling `os.path.split` on the path arg, which returns a (head,tail) tuple. If `head` is empty, the user just passed a filename and we can probably throw it in the default output dir. If head isn't empty, we can call `os.path.expandvars` on the path arg to get the path we should output to.
 
2. We should also print the full path before writing, just to be explicit with the user about what file is being created and they know exactly where to look for it when it's done.

3. We should print out something when the export has completed successfully.

setup.py
```diff
@@ -5,4 +5,5 @@
     version="0.1",
     packages=find_packages(where="src"),
     package_dir={"": "src"},
+    package_data={"delphi.static": ["*"]},
```
joshuawe, line 8: Does this line mean:
1. There is *package_data* in `delphi/static` and
2. `["*"]` states that everything in `delphi/static` should be considered part of the package_data?

joshuawe, line 8: Does this line mean:
1. There is *package_data* in `delphi/static` and
2. `["*"]` states that everything in `delphi/static` should be considered part of the package_data?

joshuawe (dismissed):
This branch is behind *main* branch, so a rebase is necessary

setup.py
menamerai: I believe so, yes. You can also specify what kind of files get considered, like `"*.csv"`, `"*.pkl"`, but I'm leaving `"*"` for now because we're not sure what kind of format we'll be using for our data.

menamerai:
@joshuawe I rebased, not sure if I did that correctly. Does it look ok?

jaidhyani (approved):

PR #43, opened 2024-02-21
Inference no copy

transcendingvictor:
Change to the inference.py module, where the pandas dataframe is created by a list comprehension of each of the rows of the `logsprobs` numpy array.
* This is needed to see the display on huggingface in batches.
* The the elements of the list comprehension are just references (not copies) to the numpy array.
* The list object itself occupies does occupy, but it's relatively very small.


scripts/inference.py
```diff
@@ -35,27 +36,22 @@ def main(
 
     model = AutoModelForCausalLM.from_pretrained(model_name)
 
-    logprobs_list = []
+    # logprobs_list = []
```
jettjaniak: remove

jettjaniak (approved):
Nice! Did you test it?

transcendingvictor:
It was tested, indeed.

PR #44, opened 2024-02-21
map tokens to hf

### LINKED ISSUE #33: token map in HF
@jaidhyani has more context

transcendingvictor:
The map_tokens.py script was change so that now it uploads the tokenmap to Hugging Face instead of saving a local pickle file.
+ The file was uploaded to HF and this is how it looks like [[link](https://huggingface.co/datasets/transcendingvictor/tinystories-v2-clean-tokenized-token-map)].
+ The dataset has two columns one for token positions and other one for token IDs.
+ Please comment (check)


scripts/map_tokens.py
```diff
@@ -1,22 +1,53 @@
...
+    )  # outputs the dictionary: dict[int, list[tuple[int, int]]]
 
-    with open(f"data/{args.output}", "wb") as f:
-        pickle.dump(mapping, file=f)
+    df_dataset = pd.DataFrame.from_dict(
+        {"token_id": list(mapping.keys()), "token_positions": list(mapping.values())}
+    )
+    df_dataset.set_index("token_id", inplace=True)  # Set the token_id as the index
+
+    hf_dataset = Dataset.from_pandas(df_dataset)
```
jettjaniak: I wonder, are there other constructors for Dataset, like from_dict, from_numpy etc.? Maybe we don't need to go through DataFrame each time.

transcendingvictor: + `complete_sorted_mapping` is the dictionary where the tokens that don't appear in the dataset are given a key (which dont have in the token_map.py file) and a `None` value.
+ QUESTION: shall we change the `map_token.py` file (specifically the initialization of the `mapping`dictso that the dictionary include all the tokens regardless of whether they appear on the dataset? (this would already give a sorted and complete dictionary for tokens from 0 to 4095)
-> I think this would be more memory efficient by allocating the memory from instantiation, but I'm not sure.

menamerai (commented):
I wonder if it's worth it to leave the local save functionality. As in it'd be useful to save the token mapping to HF directly, but retaining the ability to save locally would be nice.

scripts/map_tokens.py
jettjaniak: it makes sense to change the function in eval/token_map.py to return a list instead of a dict

jettjaniak:
> retaining the ability to save locally would be nice

what would we need this for?

scripts/map_tokens.py
```diff
@@ -1,23 +1,50 @@
...
-    mapping = token_map(dataset)
+    mapping = token_map(
+        dataset
+    )  # outputs the dictionary: dict[int, list[tuple[int, int]]]
 
-    with open(f"{STATIC_ASSETS_DIR}/{args.output}", "wb") as f:
-        pickle.dump(mapping, file=f)
+    complete_mapping = {key: mapping.get(key, None) for key in range(4096)}
+
+    hf_dataset = Dataset.from_dict({"prompt_pos_idx": list(complete_mapping.values())})
```
jaidhyani: Are we guaranteed to maintain ordering here?

jaidhyani: Seems like it might be better for this to be a list and not use a dictionary at all, since ordering is going to be deterministic and indices are ordered integers.

jaidhyani (dismissed):
The current implementation seems to reply on `.dict.values()` returning values in order.

scripts/map_tokens.py
transcendingvictor: You are right.
Dictionaries preserve order in python 3.7+.
List would also work and avoid creating a dictionary.
I just commited that.

jettjaniak (approved):
lgtm 🚀

PR #45, opened 2024-02-26
Added option to save to full paths

### LINKED ISSUE #42: Static dir follow-ups
Minor follow-up tasks for scripts writing to the static output dir #41 (`label_all_tokens.py`) 

1. We shouldn't necessarily force the output into the `STATIC_ASSETS_DIR`. Especially if the user specified a full path or a partial path with directory references, e.g.:
 
 `/home/me/Desktop/tokens.csv`
 `subdir/tokens.csv`
 
In the first case, the user has specified a full path, and we should use it. In the second, we can assume they're referring to a relative path from the current working directory, and we should output there.
 
We can get that behavior by calling `os.path.split` on the path arg, which returns a (head,tail) tuple. If `head` is empty, the user just passed a filename and we can probably throw it in the default output dir. If head isn't empty, we can call `os.path.expandvars` on the path arg to get the path we should output to.
 
2. We should also print the full path before writing, just to be explicit with the user about what file is being created and they know exactly where to look for it when it's done.

3. We should print out something when the export has completed successfully.

menamerai:
Address issue #42 

Added options to script to allow users to save to full path instead of the fixed static dir. Also modified the `scripts/map_tokens.py` file for consistency in scripts.

I noticed that the part to allow for this save option is similar across scripts. Should I make a general function for this?

### DIFF
--- a/scripts/label_all_tokens.py
+++ b/scripts/label_all_tokens.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 import pickle
 
 from tqdm.auto import tqdm
@@ -32,10 +33,14 @@ def main():
         default="delphi-suite/delphi-llama2-100k",
         required=False,
     )
+    parser.add_argument(
+        "--output",
+        help="Output path name. Must include at least output file name.",
+        default="labelled_token_ids_dict.pkl",
+    )
     args = parser.parse_args()
 
     # Access command-line arguments
-
     model_name = args.model_name
 
     print("\n", " LABEL ALL TOKENS ".center(50, "="), "\n")
@@ -86,8 +91,13 @@ def main():
         labelled_token_ids_dict[token_id] = labels[0][0]
 
     # Save the labelled tokens to a file
-    filename = "labelled_token_ids_dict.pkl"
-    filepath = STATIC_ASSETS_DIR.joinpath(filename)
+    if os.path.split(args.output)[0] == "":
+        filepath = STATIC_ASSETS_DIR.joinpath(args.output)
+        print(f"Outputting file {args.output} to path {filepath}")
+    else:
+        filepath = os.path.expandvars(args.output)
+        print(f"Outputting to path {filepath}")
+
     with open(f"{filepath}", "wb") as f:
         pickle.dump(labelled_token_ids_dict, f)
 
--- a/scripts/label_all_tokens.py
+++ b/scripts/label_all_tokens.py
@@ -34,8 +34,8 @@ def main():
         required=False,
     )
     parser.add_argument(
-        "--output",
-        help="Output path name. Must include at least output file name.",
+        "--output_path",
+        help="Output path name.",
         default="labelled_token_ids_dict.pkl",
     )
     args = parser.parse_args()
@@ -65,7 +65,7 @@ def main():
     # Save the list of all tokens to a file
     filename = "all_tokens_list.txt"
     filepath = STATIC_ASSETS_DIR.joinpath(filename)
-    with open(f"{filepath}", "w", encoding="utf-8") as f:
+    with open(str(filepath), "w", encoding="utf-8") as f:
         f.write(tokens_str)
 
     print(f"Saved the list of all tokens to:\n\t{filepath}\n")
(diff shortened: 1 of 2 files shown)


scripts/map_tokens.py
jettjaniak, line 1: I woudn't touch map_tokens, this will conflict with Victor's work and we probably don't need to save pickle anyway

scripts/label_all_tokens.py
```diff
@@ -86,8 +91,13 @@ def main():
...
-    filename = "labelled_token_ids_dict.pkl"
-    filepath = STATIC_ASSETS_DIR.joinpath(filename)
+    if os.path.split(args.output)[0] == "":
+        filepath = STATIC_ASSETS_DIR.joinpath(args.output)
+        print(f"Outputting file {args.output} to path {filepath}")
+    else:
+        filepath = os.path.expandvars(args.output)
+        print(f"Outputting to path {filepath}")
+
     with open(f"{filepath}", "wb") as f:
```
jettjaniak:
```suggestion
    with open(filepath, "wb") as f:
```

scripts/label_all_tokens.py
```diff
@@ -86,8 +91,13 @@ def main():
         labelled_token_ids_dict[token_id] = labels[0][0]
 
     # Save the labelled tokens to a file
-    filename = "labelled_token_ids_dict.pkl"
-    filepath = STATIC_ASSETS_DIR.joinpath(filename)
+    if os.path.split(args.output)[0] == "":
```
jettjaniak: .dirname would be more expressive

scripts/label_all_tokens.py
```diff
@@ -86,8 +91,13 @@ def main():
...
+        filepath = STATIC_ASSETS_DIR.joinpath(args.output)
+        print(f"Outputting file {args.output} to path {filepath}")
+    else:
+        filepath = os.path.expandvars(args.output)
```
jettjaniak: why?

scripts/label_all_tokens.py
```diff
@@ -86,8 +91,13 @@ def main():
...
+        print(f"Outputting to path {filepath}")
```
jettjaniak: why do we need two separate print statements?

scripts/label_all_tokens.py
```diff
@@ -32,10 +33,14 @@ def main():
         default="delphi-suite/delphi-llama2-100k",
         required=False,
     )
+    parser.add_argument(
+        "--output",
+        help="Output path name. Must include at least output file name.",
```
jettjaniak: I think it's either _output path_ or _output file name_, not output path name. This is a bit weird though, because: what if I want to save in my current working directory? I'll just pass "myfilename.pkl" meaning "./myfilename.pkl" but it'll save in static dir.

scripts/label_all_tokens.py
jaidhyani: e.g. `$HOME`

scripts/label_all_tokens.py
jaidhyani: I endorse both points.

On reflection, we should be *writing* to the static dir very rarely, and when we do it's reasonable to expect the user to move the files there and check them in. The main purpose of the static dir is to have a straightforward way to *read*.

So output paths should just use the path the user passes in the most straightforward way possible, without any reference to the static dir.

scripts/label_all_tokens.py
jaidhyani: Oh, sorry, think I misinterpreted this - you're asking "why" about the whole block, in which case the answer is "because that's the straightforward reading of the issue that I wrote". I should have just made the issue "use the path the user gives without any reference to the static dir" - so that's on me.

jaidhyani (changes requested):
Looks broadly good. I agree with @jettjaniak's comments. I'm sorry about the extra hassle, which is my fault - I should have just written "cut references to any particular dir" as part of the issue.

scripts/label_all_tokens.py
menamerai: `open()` can't be used on the `Traversable` type that `filepath` is. The `Traversable.open()` class method also cannot be used, as it only works for reading mode. I leave this as `open(str(filepath))` for now

scripts/label_all_tokens.py
menamerai: Earlier in the code, line 60-69, there is a section that writes down `all_tokens_list.txt`, which the script isn't giving the user a choice on where to load in the script argument. Should I leave the code as-is, letting the script saving this to `static` or give it the same treatment as the labelled token pickle by adding an extra argument?

image

joshuawe:
Hey,
adding an command line argument where to store the token_labels is already included in #40 

I would not endorse using the `STATIC_ASSETS_DIR` as output of the `label_all_tokens.py` script. The files in `STATIC_ASSETS_DIR` should shipped with the repository, and hence not custom created by the user. So I would option for a save path determined by command line argument. (Similar, to what Jai said).

Also, the first message of this PR states that `scripts/map_tokens.py` was modified, but I can only see changes in `scripts/label_all_tokens.py`.

All in all I am not sure, if this PR will now fix any issue not part of #40 ? :/

scripts/label_all_tokens.py
jettjaniak: No, I was asking just about expandvars - I haven't seen this before

menamerai:
Hey @joshuawe. Yes `scripts/map_tokens.py` was modified, but we have decided to leave the file be, as it would conflict with Victor's work, and we don't need to save the result there locally anyways.

Also, yes I agree that a command line argument to choose the save path would be great, but if it is already included in #40, then I don't think that this PR is very productive. @jaidhyani do you think I should close this PR, or is there something different you had in mind that I could work towards?

PR #31, opened 2024-02-13
Training script

jettjaniak:

PR #54, opened 2024-03-06
Training_script_refactor

jaidhyani:
This is a WIP refactor of the training script. It's not in a good state and there's a high probability I broke something while refactoring, but it has one redeeming quality: it runs. Does it run correctly and produce usable models? Excellent question, I have no idea yet.

From `src/`: ```python delphi/train/training.py```

### DIFF
--- a/src/delphi/constants.py
+++ b/src/delphi/constants.py
@@ -1,3 +1,6 @@
 from importlib.resources import files
 
 STATIC_ASSETS_DIR = files("delphi.static")
+
+CORPUS_DATASET = "delphi-suite/tinystories-v2-clean"
+TOKENIZED_CORPUS_DATASET = "delphi-suite/v0-tinystories-v2-clean-tokenized"
--- a/src/delphi/train/architectures.py
+++ b/src/delphi/train/architectures.py
@@ -1,4 +1,3 @@
-import os
 from dataclasses import fields
 
 import torch
--- a/src/delphi/train/architectures.py
+++ b/src/delphi/train/architectures.py
@@ -24,6 +24,7 @@ args_to_load_from_checkpoint = {
     ModelTypes.MAMBA: [
         "n_layers",
         "model_dim",
+        "vocab_size",
     ],
 }
 
--- a/src/delphi/train/architectures.py
+++ b/src/delphi/train/architectures.py
@@ -30,12 +30,9 @@ args_to_load_from_checkpoint = {
 
 def initialize_model(**model_args) -> torch.nn.Module:
     if model_args["architecture"] == ModelTypes.LLAMA2C:
-        llama_model_args = model_args.copy()
         # filter model_args for fields in Llama2ModelArgs
         llama2_arg_names = {f.name for f in fields(Llama2ModelArgs)}
-        llama2_args = {
-            k: v for k, v in llama_model_args.items() if k in llama2_arg_names
-        }
+        llama2_args = {k: v for k, v in model_args.items() if k in llama2_arg_names}
         return Llama2Model(Llama2ModelArgs(**llama2_args))
     else:
         raise NotImplementedError(
--- a/src/delphi/train/architectures.py
+++ b/src/delphi/train/architectures.py
@@ -1,0 +1,77 @@
+import os
+from dataclasses import fields
+
+import torch
+from llama2c import model_export
+from llama2c.model import ModelArgs as Llama2ModelArgs
+from llama2c.model import Transformer as Llama2Model
+
+
+class ModelTypes:
+    LLAMA2C = "llama2c"
+    MAMBA = "mamba"
+
+
+args_to_load_from_checkpoint = {
+    ModelTypes.LLAMA2C: [
+        "dim",
+        "n_layers",
+        "n_heads",
+        "n_kv_heads",
+        "vocab_size",
+        "multiple_of",
+        "max_seq_len",
+    ],
+    ModelTypes.MAMBA: [
+        "n_layers",
+        "model_dim",
+    ],
+}
+
+
+def initialize_model(**model_args) -> torch.nn.Module:
+    if model_args["architecture"] == ModelTypes.LLAMA2C:
+        llama_model_args = model_args.copy()
+        # filter model_args for fields in Llama2ModelArgs
+        llama2_arg_names = {f.name for f in fields(Llama2ModelArgs)}
+        llama2_args = {
+            k: v for k, v in llama_model_args.items() if k in llama2_arg_names
+        }
+        return Llama2Model(Llama2ModelArgs(**llama2_args))
+    else:
+        raise NotImplementedError(
+            f"Architecture {model_args['architecture']} not yet implemented"
+        )
+
+
+def load_model(model_args, checkpoint) -> torch.nn.Module:
+    arch = model_args["architecture"]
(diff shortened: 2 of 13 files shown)


jaidhyani:
I don't think this is totally ready yet, but at this point I think it's probably worth it to (1) merge this into the `training-script` branch and (2) then merge the `training_script` branch into `main`.

src/delphi/train/architectures.py
```diff
@@ -0,0 +1,73 @@
...
+    arch = model_args["architecture"]
+    checkpoint_model_args = checkpoint["model_args"]
+    for k in args_to_load_from_checkpoint[arch]:
+        model_args[k] = checkpoint_model_args[k]
+    if arch == ModelTypes.LLAMA2C:
+        # create the model
+        gptconf = Llama2ModelArgs(**model_args)
+        model = Llama2Model(gptconf)
+        state_dict = checkpoint["model"]
+        # fix the keys of the state dictionary :(
```
SrGonao, line 54: Are we sure this still happens?

src/delphi/train/architectures.py
```diff
@@ -0,0 +1,73 @@
...
+        "dim",
+        "n_layers",
+        "n_heads",
+        "n_kv_heads",
+        "vocab_size",
+        "multiple_of",
+        "max_seq_len",
+    ],
+    ModelTypes.MAMBA: [
+        "n_layers",
```
SrGonao, line 25: Here also vocab_size should be set

src/delphi/train/gigaconfig.py
```diff
@@ -0,0 +1,87 @@
...
+    max_epochs=2,
+    eval_interval=500,
+    eval_iters=10,
+)
+
+
+def assert_config_sanity(config: GigaConfig):
+    assert config.vocab_source in ["llama2", "custom"]
+    assert (
+        config.vocab_source == "custom" or config.vocab_size == 32000
```
SrGonao: do we want to keep 32K option or just the custom? I think just custom

src/delphi/train/iteration_params.py
```diff
@@ -0,0 +1,29 @@
...
+    num_steps: int
+    eval_iters: int
+    lr_decay_iters: int
+    tokens_per_iter: int
+
+
+def set_iteration_params(config, train_ds, validation_ds) -> IterationParams:
+    num_batches = len(train_ds) // config.batch_size
+    num_steps = num_batches // config.gradient_accumulation_steps
+    eval_iters = min(12, len(validation_ds) // config.batch_size)
```
SrGonao, line 16: what is the 12

jaidhyani: Something that was already there that I didn't want to change without understanding it first: https://github.com/delphi-suite/delphi/pull/31/files#diff-c113425bb7a4b6c38858b09ca918bc89bd243d139cdca30ab5dd386d9690935bR94

src/delphi/train/tokenized_chunks_dataset.py
```diff
@@ -0,0 +1,57 @@
...
+        ).to(self.device)
+        self.batched_tokens = tensor_tokens.flatten().unfold(
+            0, self.max_len + 1, self.max_len
+        )
+        self.indices = self._default_indices()
+
+    def _default_indices(self):
+        return list(range(len(self.batched_tokens)))
+
+    def shuffle(self, epoch: int):
```
SrGonao, line 35: I don't know if this is correct. Shuffle alters the list in place, why are we changing the indices instead of the list?

src/delphi/train/architectures.py
jaidhyani: No idea tbqh, that's just something I copied and never checked.

SrGonao (commented):
I left a couple comments, only noticeable thing is on the shuffle, the rest is minor points and then mamba stuff that I should do

src/delphi/train/tokenized_chunks_dataset.py
jaidhyani: Idempotency. I want `shuffle(x)` to result in the same state regardless of when it's called - so not dependent on the state of the shuffling before it's called. By making it a deterministic shuffle of the default indices we can guarantee that it doesn't matter when you call `shuffle(x)`, you'll still get the same result.

Technically we don't need this for reproducibility if we commit to always calling shuffle with the same arguments in the same sequence, but I prefer to have it work this way to minimize the amount of state that needs to be kept track of when debugging - I don't want to have to repeat every shuffle in a given sequence to reproduce a problem.

src/delphi/constants.py
jannik-brinkmann, line 1: seems good for now, but in the future we should consider adding the dataset as a parameter to the config

jaidhyani: Could be both - a config param that defaults to a pre-defined constant.

src/delphi/train/iteration_params.py
jannik-brinkmann: as far as I can tell this is a constant set by Karpathy to log at least every 12 batches.

src/delphi/train/tokenized_chunks_dataset.py
```diff
@@ -0,0 +1,57 @@
...
+from torch.utils.data import Dataset
+
+from delphi.train.shuffle import shuffle_list
+
+
+class TokenizedChunksDataset(Dataset):
+    def __init__(self, tokenized_docs, max_seq_len, device):
+        self.device = device
+        self.tokenized_docs = tokenized_docs
+        self.doc_len = len(tokenized_docs[0]["tokens"])
```
jannik-brinkmann, line 11: "doc_len" = "document_length" = context length?

src/delphi/train/tokenized_chunks_dataset.py
jannik-brinkmann: I like it that way

src/delphi/train/training.py
```diff
@@ -1,331 +1,62 @@
...
-        if k.startswith(unwanted_prefix):
-            state_dict[k[len(unwanted_prefix) :]] = state_dict.pop(k)
-    model.load_state_dict(state_dict)
-    iter_num = checkpoint["iter_num"]
-    best_val_loss = checkpoint["best_val_loss"]
-model.to(device)
+model_training_state = load_model_training_state(config, device)
 
+# setup eval callbacks
+eval_callbacks = [save_checkpoint_if_needed]
```
jannik-brinkmann, line 41: maybe I am missing something, but if eval_callbacks only includes saving the checkpoint, where does it log to wandb?

jannik-brinkmann: nvm, found it in line 44 lol

jannik-brinkmann (approved):
looks good to me. as discussed in Discord, I will start a training run tomorrow after the PR is merged to see whether the performance matches our current models

src/delphi/train/tokenized_chunks_dataset.py
jaidhyani: Document length, but I don't think that's context length. `max_seq_len` is model context length - or more accurately, how long training samples are.

src/delphi/train/tokenized_chunks_dataset.py
SrGonao: I just didnt see where you then apply the indices to the list, but if you do it somewhere I trust you

PR #55, opened 2024-03-10
Add function to tokenize text stories and split into batches

### LINKED ISSUE #49: dataset tokenization script
First, upload just the tokenizer to delphi-suite/tokenizer. Something like this can work
```
from transformers import AutoTokenizer
AutoTokenizer.from_pretrained("v0-llama2-...").push_to_hub("tokenizer")
```
if it doesn't, check [this](https://stackoverflow.com/questions/77265638/how-can-i-push-a-custom-tokenizer-to-huggingface-hub) out. No need to push any code to the repo for this.

Then make a script to tokenize text dataset and upload it to HF
1. pull https://huggingface.co/datasets/delphi-suite/tinystories-v2-clean (dataset name can be an argument with a default, but assume the same format)
2. tokenize each prompt, separate by EOS token and concatenate, divide into 512 length (argument) chunks and prepend BOS to each chunk - this should be doing the same things [this script](https://github.com/delphi-suite/delphi/blob/38b04996edd9febbfba189d03c22bb92c6095fcc/scripts/upload_tokens.py) is doing, but simpler and using HF tokenizer class (PretokDataset is defined in src/llama2c/tinystories.py)

siwei-li:
The test is right now two cases for text input and tokenized input, lmk if there's anything to add to it


src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,39 @@
+from collections import deque
+from typing import Union
+
+from transformers import PreTrainedTokenizerBase
+
+
+def get_tokenized_batches(
+    text_stories: Union[list[str], list[list[int]]],
```
siwei-li: core function to take in tokenized input
for the inner while loop

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,39 @@
...
+    tokenizer: PreTrainedTokenizerBase,
+    context_size: int,
+    input_tokenized=False,
+) -> list[list[int]]:
+    dq = deque()
+    samples = []
+
+    prompt_idx = 0
+    while prompt_idx < len(text_stories):
+        while len(dq) < context_size:
```
siwei-li: line 18: prompt_idx < len(text_stories)

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,39 @@
...
+            if not input_tokenized:
+                dq.extend(
+                    tokenizer.encode(text_story, add_special_tokens=False)
+                    + [tokenizer.eos_token_id]
+                )
+            else:
+                dq.extend(text_story)
+                dq.append(tokenizer.eos_token_id)
+            prompt_idx += 1
+
```
siwei-li: check the deque

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,39 @@
...
+        sample = [tokenizer.bos_token_id]
+        for i in range(context_size - 1):  # peek at and not pop the last element
+            sample.append(dq.popleft())
+        sample.append(dq[0])
+
+        samples.append(sample)
+
+    if dq:
```
siwei-li: discard the last chunk

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,39 @@
...
+                dq.extend(
+                    tokenizer.encode(text_story, add_special_tokens=False)
+                    + [tokenizer.eos_token_id]
+                )
+            else:
+                dq.extend(text_story)
+                dq.append(tokenizer.eos_token_id)
+            prompt_idx += 1
+
+        sample = [tokenizer.bos_token_id]
```
siwei-li: make a sample given a deque -> function

tests/train/test_tokenizer.py
```diff
@@ -0,0 +1,72 @@
...
+    get_tokenized_batches,
+    make_new_samples,
+)
+
+tokenizer = AutoTokenizer.from_pretrained("delphi-suite/stories-tokenizer")
+
+
+def test_extend_deque():
+    CTX_SIZE = 10
+    dataset = load_validation_dataset("delphi-suite/tinystories-v2-clean")
```
siwei-li: generate a random dataset

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,50 @@
...
+from transformers import PreTrainedTokenizerBase
+
+
+def extend_deque(
+    dq: deque[int],
+    context_size: int,
+    text_stories: list[str],
+    prompt_idx: int,
+    tokenizer: PreTrainedTokenizerBase,
+) -> int:
```
joshuawe: Please provide docstrings that explain what each function is supposed to do. If simple, a sentence is sufficient. Otherwise write as much text as you need :)

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,50 @@
...
+        text_story = text_stories[prompt_idx]
+        dq.extend(
+            tokenizer.encode(text_story, add_special_tokens=False)
+            + [tokenizer.eos_token_id]
+        )
+        prompt_idx += 1
+    return prompt_idx
+
+
+def make_new_samples(
```
joshuawe: I am not sure, if I understand this function correctly. Docstring and additional comments in the code would be super helpful.
First, what information is `dq` expected to have? I assume it contains a list of token_ids that we split into lists of fixed lengths of `context_size`, correct?

joshuawe: Of course, with a bos token at the beginning of each sample

src/delphi/train/dataset_tokenization.py
joshuawe: Do we need `prompt_idx < len(text_stories)` to not overshoot the end of `text_stories`, even if we have not reached `context_size`yet?

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,50 @@
...
+    while len(dq) >= context_size:
+        sample = [tokenizer.bos_token_id]
+        for _ in range(context_size - 1):  # peek at and not pop the last element
+            sample.append(dq.popleft())
+        sample.append(dq[0])
+        samples.append(sample)
+    return samples
+
+
+def get_tokenized_batches(
```
joshuawe: Please please a docstring 😄 
`text_stories` is a list of stories? Each story is of arbitrary length and is not tokenized? E.g.: "There once was a dog called Barky. Barky lived .... "?
And in the end we split each of the stories into chunks that are of max size `context_size`? Is that the main task of the `get_tokenized_batches` function?

tests/train/test_tokenizer.py
joshuawe: Will review all tests once you have updated the test to not load validation dataset :)

joshuawe (changes requested):
- add docstrings
- update test to not download the validation dataset from huggingface

siwei-li:
@joshuawe I added the docstrings to the functions,
the first two functions `extend_deque()` and `make_new_samples()` serve as two helper functions;

I noticed that the failing check is for the pytest in 
> tests/eval/test_token_labelling.py:2: in <module>
>     import spacy
> E   ModuleNotFoundError: No module named 'spacy'

joshuawe:
Hi @siwei-li , will check your updated code.
The error that we currently see [(Link)](https://github.com/delphi-suite/delphi/actions/runs/8319996761/job/22764114294?pr=55) is not because of the missing spacy module (I assume that your error message was from your local machine). 
Following the link we see that beartype raised a type error in the `extend_deque()`. It seems the parameter `text_stories` expected the type `list[str]` but got something like `list[int]`.

> ```Python
> E   beartype.roar.BeartypeCallHintParamViolation: Function delphi.train.dataset_tokenization.extend_deque() parameter 
> text_stories=[[1581, 1327, 61, 300, 3003, 1427, 1029, 2570, 1736, 3552, 3170, 635, 3551, 91, 3052, 2560,...]] violates type hint 
> list[str] under non-default configuration BeartypeConf(claw_is_pep526=True, is_color=None, is_debug=False, 
> is_pep484_tower=False, strategy=<BeartypeStrategy.O1: 2>, warning_cls_on_decorator_exception=<class 
> 'beartype.roar._roarwarn.BeartypeClawDecorWarning'>), as list index 11 item list [3638, 2316, 723, 505, 3953, 325, 742, 2799, 
> 3320, 4063, 159, 3931, 3300, 3072, 3636, 2544, ...] not instance of str.```

src/delphi/train/dataset_tokenization.py
joshuawe: @siwei-li , a reply to my questions would have been much appreciated, instead of just resolving them. They might be simple or even stupid questions, but it would have helped me understand your code.

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,96 @@
...
+
+    The usage of a deque here aims to save the memory as opposed to
+    load all the stories and tokenize them at once.
+
+    Args:
+        dq (deque[int]): Deque to extend with tokenized tokens.
+        context_size (int): Size of the context(input sequences).
+        text_stories (list[str]): List of (untokenized) text stories to be tokenized.
+        prompt_idx (int): Index of the current text story.
+        tokenizer (PreTrainedTokenizerBase): Tokenizer to encode the text strings.
```
jettjaniak: just a note that specifying types here is not required, we have type hints

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,96 @@
...
+        prompt_idx += 1
+    return prompt_idx
+
+
+def make_new_samples(
+    dq: deque[int], context_size: int, tokenizer: PreTrainedTokenizerBase
+) -> list[list[int]]:
+    """
+    Generates new samples for training by creating sequences of tokens
+    from the deque until the deque is empty.
```
jettjaniak: not until it's empty

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,96 @@
...
+    Returns:
+        list[list[int]]: List of token sequences of the same length(context_size).
+    """
+
+    samples = []
+    while len(dq) >= context_size:
+        sample = [tokenizer.bos_token_id]
+        for _ in range(context_size - 1):  # peek at and not pop the last element
+            sample.append(dq.popleft())
+        sample.append(dq[0])
```
jettjaniak: comment on the first line is not clear to me, and it'd be good to comment on last line

src/delphi/train/dataset_tokenization.py
jettjaniak, line 1: replace "stories" with "documents", as this is a general functionality

jettjaniak: and I guess prompt_idx should be doc_idx

src/delphi/train/dataset_tokenization.py
jettjaniak: there are no batches, so I think the function name should be changed? to something like "tokenize_dataset"

tests/train/test_tokenizer.py
```diff
@@ -0,0 +1,81 @@
...
+from transformers import AutoTokenizer
+
+from delphi.eval.utils import load_validation_dataset
+from delphi.train.dataset_tokenization import (
+    extend_deque,
+    get_tokenized_batches,
+    make_new_samples,
+)
+
+tokenizer = AutoTokenizer.from_pretrained("delphi-suite/stories-tokenizer")
```
jettjaniak:
```suggestion
```

src/delphi/train/dataset_tokenization.py
jettjaniak: this file should be delphi/dataset/tokenization.py

tests/train/test_tokenizer.py
jettjaniak, line 1: when the functions you're testing are in a delphi/x/y.py then your test file should be tests/x/test_y.py

src/delphi/train/dataset_tokenization.py
```diff
@@ -0,0 +1,96 @@
...
+        dq.extend(
+            tokenizer.encode(text_story, add_special_tokens=False)
+            + [tokenizer.eos_token_id]
+        )
+        prompt_idx += 1
+    return prompt_idx
+
+
+def make_new_samples(
+    dq: deque[int], context_size: int, tokenizer: PreTrainedTokenizerBase
```
jettjaniak: it should just take bos_token_id: int as argument, rather than tokenizer

tests/train/test_tokenizer.py
```diff
@@ -0,0 +1,81 @@
...
+        while len(dq) >= CTX_SIZE:
+            for _ in range(CTX_SIZE - 1):
+                dq.popleft()
+
+
+def test_make_new_sample():
+    for _ in range(100):
+        total_tokens = random.randint(100, 1000)
+        context_size = random.randint(5, total_tokens // 2)
+        dq = collections.deque([random.randint(3, 1000) for _ in range(total_tokens)])
```
jettjaniak: https://docs.python.org/3/library/random.html#random.choices
instead of for loop

jettjaniak:
We need a script in scripts/ that will take input dataset name, output dataset name, tokenizer name and HF credentials and tokenize and push to HF. The column name in the output dataset should be "tokens". The column name in the input dataset should be an optional argument, that defaults to the only column that is available or fails if there are more than one column. Please check (on a subset of the dataset) how slow the tokenization is. If it takes more than a few minutes, we might want to use tokenizer.batch_encode

siwei-li:
The script for HF is added @jettjaniak 
(Uploaded the tokenized 'stories' dataset to https://huggingface.co/datasets/delphi-suite/batched-tokenized-stories)

tests/dataset/test_tokenizer.py
```diff
@@ -0,0 +1,88 @@
...
+        if prompt_idx < len(text_stories) - 1:
+            # assert that the deque has grown large enough in each round
+            assert len(dq) >= CTX_SIZE
+        while len(dq) >= CTX_SIZE:
+            for _ in range(CTX_SIZE - 1):
+                dq.popleft()
+
+
+def test_make_new_sample(tokenizer):
+    for _ in range(100):
```
joshuawe, line 44: Why is this done one hundred times?

src/delphi/dataset/tokenization.py
```diff
@@ -0,0 +1,107 @@
...
+            dq.extend(input_ids + [tokenizer.eos_token_id])
+        doc_idx += batch_size
+    return doc_idx
+
+
+def make_new_samples(
+    dq: deque[int], context_size: int, bos_token_id: int
+) -> list[list[int]]:
+    """
+    Generates new samples for training by creating sequences of tokens
```
joshuawe, line 46: I find the explanation a bit confusing (or I am confused). 
This function does not generate entirely new content, correct?
It reduces the length pre-existing samples by clipping it to `context_size`, correct?
A different wording would make it easier to understand. You could also consider renaming the function `make_new_samples` to something such as `clip_samples` or similar. But this is up to you.

joshuawe: I see, my previous assumption was wrong. This function does not clip the samples, but rather splits them, prepending a BOS token as well as adding the final token from the previous split as well.

joshuawe (approved):
can be merged, if no rebase is necessary

jettjaniak (approved):

PR #56, opened 2024-03-11
Model loss difference notebook

### LINKED ISSUE #50: evals research notebook
extending #13 
add an option to sample prompts from a (range of quantiles) distribution of loss difference between two models
get a minimal working version quickly before adding any complex functionality
but write some library code when it makes sense
image

menamerai:
WIP


menamerai:
Research Notebook(1)

PR #67, opened 2024-03-20
dataclass fields have metadata! Let's use it for documentation

jaidhyani:
and automatically include it in e.g. the training script --help

Example of how this works:

In a config dataclass:
```python3
    device: str = field(
        default="auto", metadata={"help": "device to use (cuda, mps, cpu)"}
    )
 ```
 
 When you `run_training.py --help`:
 ```
  --device DEVICE       device to use (cuda, mps, cpu). Default: auto
 ```

### DIFF
--- a/src/delphi/train/config/gigaconfig.py
+++ b/src/delphi/train/config/gigaconfig.py
@@ -17,38 +17,70 @@ class GigaConfig:
     model_config: ModelConfig
     # meta
     run_name: str = datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
-    output_dir: str = os.path.join(
-        platformdirs.user_data_dir(appname="delphi"), run_name
+    output_dir: str = field(
+        default=os.path.join(platformdirs.user_data_dir(appname="delphi"), run_name),
+        metadata={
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jaidhyani** [comment]
Also I realized that the training script was not defaulting to --help when called without arguments. I thought I'd already fixed that, but it may have gotten lost in some confusion? Anyway, fixed that too.

File: src/delphi/train/config/gigaconfig.py [hunk 1]
```diff
@@ -17,38 +17,70 @@ class GigaConfig:
...
-    eval_interval: int = 2000
-    log_interval: int = 1
-    eval_iters: int = 100
-    always_save_checkpoint: bool = (
-        False  # if True, always save a checkpoint after each eval
+    eval_interval: int = field(default=2000, metadata={"help": "eval every N iters"})
+    log_interval: int = field(default=1, metadata={"help": "log every N iters"})
+    eval_iters: int = field(default=100, metadata={"help": "use N iters for each eval"})
+    always_save_checkpoint: bool = field(
+        default=False,  # if True, always save a checkpoint after each eval
```
jettjaniak: redundant comment

jettjaniak (approved):
neat!

PR #69, opened 2024-03-20
50 evals research notebook (sampling method)

### LINKED ISSUE #50: evals research notebook

menamerai:


menamerai:
Almost done, will add the visualization function and interactive part Thursday evening - Friday morning this week

src/delphi/eval/utils.py
```diff
@@ -115,3 +116,15 @@ def load_logprob_datasets(split: str = "validation") -> dict[str, list[list[floa
...
         for model in constants.LLAMA2_MODELS
     }
+
+
+def dict_filter_quantile(
+    d: dict[Any, float], q_start: float, q_end: float
+) -> dict[Any, float]:
+    if not (0 <= q_start < q_end <= 1):
+        raise ValueError("Invalid quantile range")
+    q_start_val = np.nanquantile(list(d.values()), q_start)
```
menamerai, line 129: I wanted to use `torch.quantile` here, but that function expects a Tensor, and Tensor expect a list of some sort, and thus mean d.values have to be casted to a list... this ends up being something like `torch.quantile(torch.Tensor(list(d.values)), q_start)`. I don't know if there is a better way of doing this

jettjaniak (approved):

PR #70, opened 2024-03-20
Support for all transformers library CausalLM models

jaidhyani:


jettjaniak (approved):
If you feel good about it let's just merge and discuss everything on Monday

PR #74, opened 2024-03-23
Make training more legible

jaidhyani:
"I have made this longer than usual because I have not had time to make it shorter." - Blaise Pascal


jaidhyani:
This is a combination of increased documentation, minor cleanup and refactoring, and removing stuff that I don't think is getting used

jettjaniak (approved):

PR #94, opened 2024-03-30
use custom HF namespaces

### LINKED ISSUE #93: tokenize dataset script with custom huggingface namespaces
Currently we can only upload to the delphi-suite namespace on HF, but ideally we would like to do that for any name space specified by the user.
Small change required in line 75

https://github.com/delphi-suite/delphi/blob/5b7ec89061c6d111c665678a97d895efe5414a53/scripts/tokenize_dataset.py#L74-L78

joshuawe:
For users to use their custom datasets for downloading, tokenizers for tokenizing and custom place for uploading we need to allow any namespace (and not limit users to "delphi-suite")

### DIFF
--- a/scripts/tokenize_dataset.py
+++ b/scripts/tokenize_dataset.py
@@ -50,8 +50,8 @@ if __name__ == "__main__":
     )
     args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


siwei-li (approved):
Looks good to me, thanks for the change!

PR #95, opened 2024-03-30
Train rework 2

### LINKED ISSUE #77: checkpoint parameterization
Train checkpointing should be parameterized on 'every n steps' as well as a specific (user-supplied) list of steps to checkpoint on (main use case is high frequency checkpoint early in training to study how the model develops early on)

checkpointing should *never* be a function of validation performance, should be based entirely on configuration; you should know exactly how many checkpoints will be created before training starts, and where they will be

### LINKED ISSUE #83: resume_from_dir should be separate argument
distinct fro output_dir

### LINKED ISSUE #84: huggingface upload logic
Checkpoints should be saved to their own branches on huggingface repos, should not include optimizer weights

jaidhyani:


SrGonao (commented):
I think I'm able to follow the code better now.

jettjaniak (approved):

PR #99, opened 2024-03-31
Add "checkpoint_mode" kwarg to plotting

### LINKED ISSUE #87: make eval plot for model checkpoints
Make a variant of plotting function that works for a high number of data points:
Instead of labeling each model on x axis, have a time scale in place,
and instead of separate error bars on each datapoint have a solid line for mean and shaded regions for quantiles.

image

siwei-li:


jettjaniak (changes requested):
notebook is not running
```
BeartypeCallHintReturnViolation: Function delphi.eval.vis_per_token_model.visualize_per_token_category.get_plot_values() return (defaultdict(None, {'nouns': (1.370191698609721, 0.4339272244650738, 0.9698742185740588),...)})) violates type hint tuple[numpy.ndarray, numpy.ndarray, numpy.ndarray] under non-default configuration BeartypeConf(claw_is_pep526=True, is_color=None, is_debug=False, is_pep484_tower=False, strategy=<BeartypeStrategy.O1: 2>, warning_cls_on_decorator_exception=<class 'beartype.roar._roarwarn.BeartypeClawDecorWarning'>), as tuple index 0 item <class "collections.defaultdict"> "defaultdict(None, {'nouns': (1.370191698609721, 0.4339272244650738, 0.9698742185740588),...)})" not instance of <protocol "numpy.ndarray">.
```

jettjaniak:
- The idea behind **kwargs was that we may be able to pass some additional arguments to plotly calls without specifying them directly, like plt.line(data, **kwargs).
- I can see it's hard to make it work that way because we have multiple plotly calls
- kwargs.get(arg_name, default) doesn't make much sense, we'd be better off just defining arg_name as an argument to the function
- but if we can't make it simple and configurable, let's just make it simple and non-configurable. so remove the kwargs and add checkpoint_mode argument

jettjaniak:
also, please remove the commented code

jettjaniak (approved):

PR #100, opened 2024-04-01
setup + requirements deps in pyproject.toml; transformers to 4.39.2

### LINKED ISSUE #58: Add requirements to setup.py
Right now pip-installing delphi is broken because it doesn't pull in dependencies. We need to specify runtime dependencies in setup.py (this will probably be mostly copying from requirements.txt)

### LINKED ISSUE #71: Update transformers to 4.39
When it comes out

jaidhyani:
This makes it possible to install delphi using pip without needing to install requirements separately. Requirements are now specified in pyproject.toml, and these are used by both pip and requirements.txt.


jaidhyani:
Tested with:

`pip install git+https://github.com/delphi-suite/delphi.git@dependencies_update`

jaidhyani:
Tested the optional cuda packages with:

`pip install "git+https://github.com/delphi-suite/delphi.git@dependencies_update#egg=delphi[cuda]"` (this is apparently the simplest syntax for installing an optional set of extra requirements from a package over git)

Fails as expected on mac...which means it's working.

pyproject.toml
```diff
@@ -1,6 +1,41 @@
...
+    "spacy==3.7.2",
+    "pandas==1.3.4",
+    "dacite==1.8.1",
+    "panel==1.4.0",
+    "jupyter_bokeh==4.0.1",
+    "transformers==4.39.2",
+]
+
+[project.optional-dependencies]
+cuda = [
```
jettjaniak, line 30: should we call this mamba_cuda?

requirements.txt
jettjaniak, line 1: what's the use-case for keeping the requirements.txt?

jettjaniak (approved):
brilliant!

requirements.txt
jaidhyani: Developer expectations - lots of people have "check out repo, make venv, install requirements.txt" muscle memory. Best to meet that expectation, or have it fail in a way that makes it obvious why it's failing (no CUDA).

pyproject.toml
jaidhyani: Yeah, that's a more accurate name. I'm going to be traveling through tomorrow, but I'll make the change in this or a subsequent PR when I get a chance

PR #103, opened 2024-04-01
tokenizer training script

jettjaniak:


scripts/train_tokenizer.py
```diff
@@ -0,0 +1,131 @@
...
+        pad_token="<pad>",
+        pad_token_id=3,
+        padding_side="right",
+    )
+    os.remove(bpe_tokenizer_file_path)
+    print("Converted tokenizer to huggingface tokenizer.")
+    
+    # push tokenizer to the hub
+    tokenizer.push_to_hub(
+        repo_id="jbrinkma/tokenizer_test",
```
jettjaniak: should use arguments

scripts/train_tokenizer.py
```diff
@@ -0,0 +1,131 @@
+import argparse
+import os
+
+from datasets import load_dataset
+from tqdm.auto import tqdm
+from transformers import LlamaTokenizerFast
+from tokenizers import SentencePieceBPETokenizer
+from tqdm import trange
```
jettjaniak:
```suggestion
from tqdm.auto import tqdm, trange
from transformers import LlamaTokenizerFast
from tokenizers import SentencePieceBPETokenizer
```

scripts/train_tokenizer.py
```diff
@@ -0,0 +1,131 @@
...
+def main(
+    vocab_size: int,
+    dataset_name: str,
+    train_size: float,
+    username: str,
+    token: str,
+    funct_test: bool = False,
+):
+    """
+    Trains a SentencePiece tokenizer on a dataset.
```
jettjaniak: What's the difference between SentencePiece and LlamaTokenizerFast?

scripts/train_tokenizer.py
```diff
@@ -0,0 +1,131 @@
...
+    Args:
+    - vocab_size: The vocabulary size of the resulting tokenizer
+    - dataset_name: The name of the dataset from which validation set will be loaded
+    - train_size: The amount of the dataset that should be used for training
+    - username: Hugging Face API username
+    - token: Hugging Face API token
+    """
+    train_ds = load_dataset(dataset_name)["train"]
+    if train_size < 1.0:
+        train_ds = train_ds.train_test_split(train_size=train_size)["train"]
```
jettjaniak: This is randomized, so not good for reproducibility. Can we just take str split argument that defaults to "train"? If you want to train on a subset, you can then specify "train[:1000]" or "train[:10%]"

scripts/train_tokenizer.py
```diff
@@ -0,0 +1,131 @@
...
+    tokenizer_model_path = get_tokenizer_model_path(
+        vocab_size=vocab_size,
+    )
+    if not os.path.isfile(tokenizer_model_path):
+        train_vocab(
+            vocab_size=vocab_size,
+            dataset=train_ds,
+        )
+        
+    import sentencepiece as spm
```
jettjaniak: to the top

scripts/train_tokenizer.py
```diff
@@ -0,0 +1,131 @@
...
+    - username: Hugging Face API username
+    - token: Hugging Face API token
+    """
+    train_ds = load_dataset(dataset_name)["train"]
+    if train_size < 1.0:
+        train_ds = train_ds.train_test_split(train_size=train_size)["train"]
+        
+    tokenizer_model_path = get_tokenizer_model_path(
+        vocab_size=vocab_size,
+    )
```
jettjaniak: I don't think this should be a separate function

scripts/train_tokenizer.py
```diff
@@ -0,0 +1,131 @@
...
+    if not os.path.isfile(tokenizer_model_path):
+        train_vocab(
+            vocab_size=vocab_size,
+            dataset=train_ds,
+        )
+        
+    import sentencepiece as spm
+    tokenizer_model_path = get_tokenizer_model_path(vocab_size=vocab_size)
```
jettjaniak: this is called twice

src/delphi/train/tokenizer.py
```diff
@@ -0,0 +1,65 @@
...
+    assert vocab_size > 0, "Vocab size must be positive"
+    
+    if not os.path.exists(cache_dir):
+        os.makedirs(cache_dir)
+
+    # export text as a single text file
+    text_file = os.path.join(cache_dir, "text.txt")
+    with open(text_file, 'w', encoding='utf-8') as file:
+        for item in dataset:
+            text = item['story']
```
jettjaniak: feature needs to be configurable. If not set, should default to the only feature there is, or fail if there are more than two

scripts/train_tokenizer.py
```diff
@@ -0,0 +1,131 @@
...
+    vocab = {sp_model.id_to_piece(index): index for index in trange(sp_model.GetPieceSize())}
+    merges = []
+    for piece_l in tqdm(vocab.keys(), total=sp_model.GetPieceSize()):
+        for piece_r in vocab.keys():
+            merge = f"{piece_l}{piece_r}"
+            piece_id = vocab.get(merge, None)
+            if piece_id:
+                merges += [(piece_l, piece_r, piece_id)]
+    merges = sorted(merges, key=lambda val: val[2])
+    merges = [(val[0], val[1]) for val in merges]
```
jettjaniak: any idea what is this doing?

jettjaniak (commented):
Thanks, looks good!
There is a lot of files juggling that I don't think we need. Let's use tempfile module to deal with that and remove the files automatically after we're done.

Claude says:

In Python, you can use the `tempfile` module to create temporary files. The `tempfile` module provides a secure and convenient way to create temporary files and directories that are automatically deleted when they are no longer needed.

Here are a few ways to create temporary files using the `tempfile` module:

1. Using `tempfile.NamedTemporaryFile()`:
```python
import tempfile

with tempfile.NamedTemporaryFile(mode='w+') as temp_file:
    temp_file.write('Some temporary data')
    temp_file.flush()
    # Do something with the temporary file
    # The file will be automatically deleted when the 'with' block ends
```

2. Using `tempfile.TemporaryFile()`:
```python
import tempfile

with tempfile.TemporaryFile(mode='w+') as temp_file:
    temp_file.write('Some temporary data')
    temp_file.seek(0)  # Move the file pointer to the beginning
    # Do something with the temporary file
    # The file will be automatically deleted when the 'with' block ends
```

3. Using `tempfile.mkstemp()`:
```python
import tempfile
import os

fd, temp_path = tempfile.mkstemp()
try:
    with os.fdopen(fd, 'w') as temp_file:
        temp_file.write('Some temporary data')
    # Do something with the temporary file
finally:
    os.remove(temp_path)  # Explicitly delete the temporary file
```

In the first two examples, the temporary files are created using a context manager (`with` block), which ensures that the files are automatically deleted when the block ends, even if an exception occurs.

In the third example, `tempfile.mkstemp()` returns a file descriptor and the path to the temporary file. You need to explicitly delete the temporary file using `os.remove()` when you're done with it.

By default, the temporary files are created in the default temporary directory of the operating system. You can specify a different directory by passing the `dir` parameter to the `tempfile` functions.

Using the `tempfile` module is recommended for creating temporary files because it handles the file creation and deletion securely and avoids common pitfalls associated with manual temporary file management.

jettjaniak:
Looks like you need to setup black, CI is failing on that

jettjaniak:
Also, could you think about some localized unit tests? Like we have a pre-defined string as a text to train on and we check if the resulting tokenizer has the same vocab and tokenizes text the way we expect.

scripts/train_tokenizer.py
jaidhyani: `test_train_split` takes an optional `seed` argument

scripts/train_tokenizer.py
jannik-brinkmann: Yes. Other scripts use a fixed, e.g. `repo_id = f"{args.username}/v0-token-map"` (in map_tokens.py), but I think a specific parameter for the repo_id might make more sense. I will add one but please let me know if you prefer it the other way.

scripts/train_tokenizer.py
jannik-brinkmann: LlamaTokenizer is just a simple wrapper for SentencePiece with the option to add `<eos>` and `<bos>` tokens (see https://github.com/meta-llama/llama/blob/main/llama/tokenizer.py). ...Fast tokenisers are functionally the same but implemented in Rust and presumably are significantly faster (although I never tested this tbh).

scripts/train_tokenizer.py
jannik-brinkmann: I think I prefer the seed option, but also like the absolute selection option. I will add a seed parameter and remove the type hint for the train_size argument, as `train_test_split` allows both int and float for absolute and relative selections, respectively.

scripts/train_tokenizer.py
jannik-brinkmann: This generates the components HF needs to initialise the SentencePieceBPETokenizer. `vocab` are all tokens, and `merges` are all combinations of tokens that exist as another token in the vocabulary (e.g. "hel" + "lo" = "hello")

scripts/train_tokenizer.py
jannik-brinkmann: regarding the use of `tempfile`: previously, I created three local files
1. the text file for training the tokenizer: 
2. the original sentencepiecetokenizer
3. when converting the tokenisers. 

I removed (1) (3), but eliminating (2) doesn't seem possible: `SentencePieceTrainer` seems to always create a local file which has to be handled (https://snyk.io/advisor/python/sentencepiece/functions/sentencepiece.SentencePieceTrainer).

scripts/train_tokenizer.py
jannik-brinkmann: removed

src/delphi/train/tokenizer.py
jannik-brinkmann: added as a parameter

PR #108, opened 2024-04-03
load config files in order, later overrides earlier, lots more testing

### LINKED ISSUE #88: move any non-trivial logic related to scripts into the main library and test it

### LINKED ISSUE #102: configs should be prioritized in argument order (later overrides former)

jaidhyani:
The main functional impact of this PR is to kill the "priority" config arg and just load configs in the order in which they're passed in to the training script, with the later overriding the former. Alongside this we do some refactoring to support testing to ensure that everything works as expected. We also improve type-casting override argument strings to better match the actual config types and make it significantly less dangerous.


jettjaniak:
why the merge commit?

jaidhyani:
> why the merge commit?

Default behavior in the github UI, clicked without thinking. I think we can configure this to default to rebase/ff.

image

jettjaniak:
https://github.com/orgs/community/discussions/12032
not configurable unfortunately :/

.vscode/launch.json
jettjaniak, line 1: let's add it to gitignore

scripts/run_training.py
```diff
@@ -52,6 +55,8 @@ def setup_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(description="Train a delphi model")
     parser.add_argument(
         "--config_files",
+        "--config_file",
+        "-c",
```
jettjaniak, line 59:
```suggestion
        "--config",
        "-c",
```
I don't care about backward compatibility, no one is using this yet

scripts/run_training.py
```diff
@@ -78,20 +83,7 @@ def setup_parser() -> argparse.ArgumentParser:
 def overrides_to_dict(overrides: list[str]) -> dict[str, Any]:
     # {"--overrides a.b.c=4 foo=false} to {"a": {"b": {"c": 4}}, "foo": False}
```
jettjaniak: that needs updating

src/delphi/constants.py
```diff
@@ -1,7 +1,9 @@
 from importlib.resources import files
+from pathlib import Path
+from typing import cast
 
 STATIC_ASSETS_DIR = files("delphi.static")
-CONFIG_PRESETS_DIR = STATIC_ASSETS_DIR / "configs"
+CONFIG_PRESETS_DIR = cast(Path, STATIC_ASSETS_DIR / "configs")
```
jettjaniak, line 6: Is this a Path?

src/delphi/train/config/utils.py
```diff
@@ -1,10 +1,12 @@
...
 import os
 from dataclasses import fields, is_dataclass
 from datetime import datetime
 from pathlib import Path
-from typing import Type
+from typing import _GenericAlias  # type: ignore
+from typing import Any, Type, TypeVar, Union
 
 import platformdirs
 from beartype.typing import Any, Iterable
```
jettjaniak:
```suggestion
from collections.abc import Iterable
```

jettjaniak (approved):
nice! 🚀 
- why do we need `filter_config_to_actual_config_values`?
- in general I'm quite overwhelmed by train/config/utils.py, I didn't realize we'll need all of this to have the functionality I imagined
- could you add some docstrings? like what are the functions doing and why we need them

scripts/run_training.py
jettjaniak: I mean, what's an example input?

src/delphi/constants.py
jaidhyani: It is now. More generally, it is a Path in practice, and doing this casting here (1) breaks nothing and (2) makes some basic operations more straightforward with less boilerplate.

jaidhyani:
> * filter_config_to_actual_config_values

I think we actually don't need this anymore, going to optimistically delete.

jaidhyani:
> * 
> * in general I'm quite overwhelmed by train/config/utils.py, I didn't realize we'll need all of this to have the functionality I imagined
> * could you add some docstrings? like what are the functions doing and why we need them

I added some more docstrings explaining how some things are used, why we need them, and how they work. I also deleted a few things that we don't need anymore.

PR #113, opened 2024-04-09
llama2 & mamba training configs

SrGonao:
(I messed up with the name of this branch. I tried renaming it but it didn't update)
I added a default mamba and llama 2 base (they are actually the same) and then the different llama 2 and mamba sizes.


scripts/training_config_examples/llama2/llama100k.json
```diff
@@ -0,0 +1,24 @@
+{
+  "model_config": {
+    "model_class": "LlamaForCausalLM",
+    "attention_bias": false,
+    "attention_dropout": 0.0,
+    "bos_token_id": -1,
+    "eos_token_id": -2,
```
jettjaniak: We probably should give the actual values from our tokenizer?

scripts/training_config_examples/llama2/llama100k.json
```diff
@@ -0,0 +1,24 @@
...
+    "hidden_act": "silu",
+    "hidden_size": 24,
+    "initializer_range": 0.02,
+    "intermediate_size": 64,
+    "max_position_embeddings": 512,
+    "num_attention_heads": 2,
+    "num_hidden_layers": 1,
+    "num_key_value_heads": 1,
+    "pretraining_tp": 1,
```
jettjaniak: what is this?

scripts/training_config_examples/llama2/llama100k.json
```diff
@@ -0,0 +1,24 @@
...
+    "rms_norm_eps": 1e-06,
+    "rope_scaling": null,
+    "rope_theta": 10000.0,
+    "tie_word_embeddings": false,
```
jettjaniak: wait, this is not a standard setting for llama, is it?

scripts/training_config_examples/mamba/mamba100k.json
```diff
@@ -0,0 +1,29 @@
...
+    "hidden_size": 24,
+    "state_size": 16,
+    "num_hidden_layers": 2,
+    "conv_kernel": 4,
+    "expand": 2,
+    "use_bias": false,
+    "use_conv_bias": true,
+    "bos_token_id": 0,
+    "eos_token_id": 0,
+    "pad_token_id": 0,
```
jettjaniak: specify actual values, except we don't have pad token

src/llama2c
jettjaniak, line 1: you somehow added a submodule that shouldn't be here

scripts/training_run.sh
jettjaniak, line 1: we probably don't need this script in the repo

scripts/training_config_examples/mamba/stories-mamba.json
```diff
@@ -0,0 +1,31 @@
...
+        "log": false,
+        "project": "delphi",
+        "entity": "set_wandb.entity_to_your_wandb_username_to_make_wandb_logging_work"
+    },
+    "batch_size": 64,
+    "max_seq_len": 512,
+    "max_epochs": 10,
+    "grad_clip": 1.0,
+    "optimizer": {
+        "gradient_accumulation_steps": 4,
```
jettjaniak: why?

scripts/training_config_examples/mamba/stories-mamba.json
```diff
@@ -0,0 +1,31 @@
+{
+    "run_name": "run",
+    "output_dir": "output",
```
jettjaniak: these should probably just take default values unless overridden via CLI argument?

scripts/training_config_examples/mamba/stories-mamba.json
```diff
@@ -0,0 +1,31 @@
...
+    "log_interval": 1,
+    "eval_iters": 100,
+    "eval_only": false,
+    "always_save_checkpoint": false,
+    "init_from": "scratch",
+    "wandb_config": {
+        "log": false,
+        "project": "delphi",
+        "entity": "set_wandb.entity_to_your_wandb_username_to_make_wandb_logging_work"
+    },
```
jettjaniak: Can we just not specify it? Not sure what the right approach here is

scripts/training_config_examples/mamba/mamba50M.json
```diff
@@ -0,0 +1,29 @@
+{
+  "model_config": {
+    "CUBLAS_WORKSPACE_CONFIG=:4096:8": "MambaForCausalLM",
```
jettjaniak: ? :D

jettjaniak (changes requested):
Thank you!

**location**
for now configs should go to `src/delphi/static/configs/stories`

**naming**
we should either go with the full name that will end up on HF so
`stories-{llama2,mamba}-[num]{m,k}.json`
or just
`[num]{m,k}.json`
please note it should be `llama2` not `llama` and `m` not `M`

**inheritance**
you can specify all of the keys and values for `model_config` that are shared in the base configs, so e.g. `llama100k.json` should not specify `"model_class": "LlamaForCausalLM"`

@jaidhyani can we easily write a test that will check if all configs in the repo properly initialize?

@jannik-brinkmann please check if llama2 configs are correct

scripts/training_config_examples/llama2/llama100k.json
SrGonao: This was there from the Jai config, I just copy pasted. Maybe @jannik-brinkmann knows?

scripts/training_config_examples/llama2/llama100k.json
SrGonao: the first 3 is how you define the specific parameters that we are using. I'm not sure I follow your question

scripts/training_config_examples/mamba/stories-mamba.json
SrGonao: I'm of the opinion that we want to have in the config all options that could be changed, but maybe this is the wrong way to view it.

scripts/training_config_examples/mamba/stories-mamba.json
SrGonao: Reposting from above: "I'm of the opinion that we want to have in the config all options that could be changed, but maybe this is the wrong way to view it"

SrGonao:
@jettjaniak, about inheritance, I'm not being able to have it working.
eg. if I have:
```
"model_config": {
        "model_class": "MambaForCausalLM",
        "vocab_size": 4096,
        "conv_kernel": 4,
        "expand": 2,
        "use_bias": false,
        "use_conv_bias": true,
        "hidden_act": "silu",
        "initializer_range": 0.1,
        "residual_in_fp32": true,
        "tie_word_embeddings": true
    }
```

in one mamba.json file and:
```"model_config": {
    "hidden_size": 952,
    "num_hidden_layers": 8
  }
```
in a 50M.json file
 and I run
`python scripts/run_training.py --config_files mamba.json --config_files 50M.json`
I get `KeyError: 'model_class'`

Could you verify it @jaidhyani ? I have changed all files to be like that either way

jaidhyani:
--config_files should be passed exactly once, with space-separated arguments if you want to pass multiple configs

scripts/training_config_examples/llama2/llama100k.json
jettjaniak: pretraining_tp

scripts/training_config_examples/llama2/llama100k.json
jettjaniak: still relevant

scripts/training_config_examples/llama2/llama100k.json
jettjaniak: I think this is wrong

scripts/training_config_examples/mamba/mamba100k.json
jettjaniak: OK, now we're not setting bos/eos token id anywhere for mambas. Won't it cause any issues when using the models?

scripts/training_config_examples/llama2/llama100k.json
jettjaniak: or at least we were training with tied embeddings previously, and we probably computed the number of parameters assuming tied embeddings

jettjaniak:
TODOs
- [x] set all the values instead of relying on defaults (we can find them here https://github.com/huggingface/transformers/tree/v4.39.3/src/transformers/models )
- [x] decide on tie_word_embeddings
- [ ] verify number of parameters by initializing the models (maybe in a test?)
- [x] fix existing tests so they can access configs in new location
- [ ] make tests that will validate the configs
- [x] figure out what to do with wandb_config, run_name, output_dir - should they be part of the config?

jaidhyani:
> TODOs
> 
> * [ ]  fix existing tests so they can access configs in new location

AFAIK there's not really a good way to include static files from outside of a package's root directory, by design. Basically all documentation on the subject assumes that static files are included in the project root - e.g. `src/` for us. I've just spent some time messing around with ways to get around this with MANIFEST.in or different setup.py options, couldn't get anything working.

jettjaniak:
can't we just determine the test file path and navigate from there to repo root?

jettjaniak:
oh, but that won't work if it's installed instead of cloned

scripts/training_config_examples/mamba/stories-mamba.json
jettjaniak: for now have a reasonable example config here that's commented

scripts/training_config_examples/mamba/stories-mamba.json
jettjaniak: these don't need to be defined, have sane defaults

jettjaniak:
gradient_accumulation_step should default to 1 in the code ❗

jettjaniak:
we need to look at all of these

```
    "run_name": " ",
    "output_dir": " ",
    "device": "auto",
    "eval_interval": 2000,
    "log_interval": 1,
    "eval_iters": 100,
    "eval_only": false,
    "always_save_checkpoint": false,
    "init_from": "scratch",
    "wandb_config": {
        "log": false,
        "project": " ",
        "entity": " "
    },
```
and change if needed

jettjaniak:
all of the model and training config were discussed and updated

jaidhyani:
#116 is a PR on top of this one to migrate static assets to the new top-level non-package path.

jettjaniak:
we need to find reasonable values for these
```
    "checkpoint_interval": -1000,
    "extra_checkpoint_iters": [
        -1000,
        -2000
    ],
    "log_interval": -1000,
```

PR #128, opened 2024-04-25
HF revamp

jettjaniak:
- we never pass tokens as argument, it needs to be set in env variable HF_TOKEN
- training config needs `out_repo_id` - set to empty string if you don't want to upload to hf
  - if this is provided, every checkpoint and final model will be uploaded to HF
- training config has `save_optimizer: bool = True`
- changed folder structure for saved models to be easily loadable from HF


SrGonao (commented):
Ok

jaidhyani (approved):
lgtm

PR #139, opened 2024-05-15
handle CUBLAS_WORKSPACE_CONFIG env var

### LINKED ISSUE #131: `CUBLAS_WORKSPACE_CONFIG`
image

jettjaniak:


SrGonao (commented):
My problem with this is that there will be a warning and then there will be a crash - torch just refuses to proceed if the correct value is not set- and so people will just see too warnings. I don't know if the best thing would just be to set it to the expected value?

jettjaniak:
you're right, fixed!

PR #142, opened 2024-05-15
add readme_path to config for HF model cards

### LINKED ISSUE #129: provide readme and model card when uploading models
image
:(

jettjaniak:
a stub of what we could use

```makrdown
---
language:
- en
license: apache-2.0
datasets:
- delphi-suite/stories
library_name: delphi
---

This model was trained using [Delphi](https://github.com/delphi-suite/delphi). See `training_config.json` and `run_context.json` for details.
```

### DIFF
--- a/src/delphi/train/config/training_config.py
+++ b/src/delphi/train/config/training_config.py
@@ -92,6 +92,10 @@ class TrainingConfig:
     # third party
     wandb: Optional[WandbConfig] = None
     out_repo_id: str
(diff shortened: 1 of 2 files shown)


SrGonao:
Are you also adding to this PR the actual way to produce the model cards or will that be a follow up? otherwise looks good

jettjaniak:
TODOs

- [ ] upload readme only to the main branch
- [ ] provide a default

PR #143, opened 2024-05-19
revamp RunContext and add gpu_name

### LINKED ISSUE #141: run_context.json should register the GPU used
it just says "cuda"

jettjaniak:


SrGonao (commented):
Looks cleaner, seems good

PR #144, opened 2024-05-20
fix problem with attention_size

jettjaniak:
some sizes have intermediate_size = 2 * hidden_size, what's that about? (should be closer to 8/3)


jettjaniak:
```ValueError: hidden_size must be divisible by num_heads (got `hidden_size`: 54 and `num_heads`: 4)```

